PR #22, opened 2022-07-05
feature/OC-02/results-multi-display

remko48:


pwa/src/templates/componentResult/ComponentResultsTemplate.module.css
```diff
@@ -0,0 +1,10 @@
+.contentRow > :last-child {
+  display: flex;
+  justify-content: flex-end;
+}
+
+.componentsGrid {
+  display: grid;
+  grid-gap: var(--nlportal-space-block-md);
+  grid-template-columns: 1fr 1fr;
+}
```
lencodes, line 10: This css rules themselves are fine. However, its unclear to what sub-component they're referring. 

e.g.: the `.contentRow` is targeting the table, but that's not clear within the context of this css.

A solution would be to add some comments to this css file, like so:

```
/*  Table */
... all table rules

/* Cards */
... all cards rules

/* Layers */
... all layer rules
```

pwa/src/templates/componentResult/ComponentResultsTemplate.tsx
```diff
@@ -0,0 +1,107 @@
...
+import { RichContentCard, Tag } from "@conduction/components";
+import { Card, Link } from "@gemeente-denhaag/components-react";
+import { HamburgerIcon, HouseIcon } from "@gemeente-denhaag/icons";
+import { Table, TableBody, TableCell, TableHeader, TableRow } from "@gemeente-denhaag/table";
+import { useTranslation } from "react-i18next";
+import { navigate } from "gatsby";
+import { ArrowRightIcon } from "@gemeente-denhaag/icons";
+
+interface ComponentsTemplateProps {
+  type: string;
```
lencodes, line 13: We could expand on this `type` prop, like so:


```
export type TComponentResultsLayout = "table" | "cards" | "layer";

interface ComponentsTemplateProps {
  type: TComponentResultsLayout;
  results: any[];
}
```

This way, when implementing this component, we cannot set a type (`string`) that does not exist.

pwa/src/templates/componentResult/ComponentResultsTemplate.tsx
```diff
@@ -0,0 +1,107 @@
...
+import * as _ from "lodash";
+import { RichContentCard, Tag } from "@conduction/components";
+import { Card, Link } from "@gemeente-denhaag/components-react";
+import { HamburgerIcon, HouseIcon } from "@gemeente-denhaag/icons";
+import { Table, TableBody, TableCell, TableHeader, TableRow } from "@gemeente-denhaag/table";
+import { useTranslation } from "react-i18next";
+import { navigate } from "gatsby";
+import { ArrowRightIcon } from "@gemeente-denhaag/icons";
+
+interface ComponentsTemplateProps {
```
lencodes, line 12: `ComponentResultsTemplateProps`*

pwa/src/templates/componentResult/ComponentResultsTemplate.tsx
lencodes: Note: we export the type so that we can import it where we're implementing this component, see other comments.

pwa/src/templates/componentResult/ComponentResultsTemplate.tsx
```diff
@@ -0,0 +1,107 @@
...
+              key={component.id}
+              title={component.name}
+              date={new Date(component.releaseDate)}
+              subTitle={component.intendedAudience}
+              onClick={() => navigate(`/components/${component.id}`)}
+              variant="case"
+            />
+          ))}
+        </div>
+      );
```
lencodes, line 104: This works great. However, I think with a small adjustment, we can manage it much more easily.

I would create three new components **within** this component file. So completely outside of your first component declaration (so after the final `};` of this file), I'd make three new components, like so:

```
const Table: React.FC = () => {
  ... all table code here
}

const Cards: React.FC = () => {
  ... all table code here
}

const Layers: React.FC = () => {
  ... all table code here
}
```

Thereafter, you can just initiate these components in your switch, like so:

```
...
case "table":
      return <Table {...{components}} />
...
```

pwa/src/templates/components/ComponentsTemplate.module.css
```diff
@@ -18,8 +20,11 @@
   margin-inline-end: var(--nlportal-space-block-xl);
 }
 
-.componentsGrid {
-  display: grid;
-  grid-gap: var(--nlportal-space-block-md);
-  grid-template-columns: 1fr 1fr;
+.buttongroup {
```
lencodes, line 23: `.buttonGroup`*

Also, maybe we could make it a bit more declarative. `layoutSwitchButtonsContainer`? I'm not sure.

pwa/src/templates/components/ComponentsTemplate.module.css
```diff
@@ -18,8 +20,11 @@
...
+  display: flex;
+  align-items: center;
+}
+
+.buttongroup > button:not(:last-child) {
```
lencodes, line 28: Might not be possible with the way we import the buttons, but generally speaking we'd like to avoid calling elements directly (so instead of `button` we would want to call `.buttonClassName`).

pwa/src/templates/components/ComponentsTemplate.tsx
```diff
@@ -107,31 +130,16 @@ export const ComponentsTemplate: React.FC<ComponentsTemplateProps> = ({ defaultT
...
+      {!getComponents.isLoading ? (
+        <ComponentResultTemplate results={filteredComponents} type={display} />
+      ) : (
+        <>
+          <Skeleton height="250px" />
+          <Skeleton height="250px" />
+        </>
+      )}
+
+      {!filteredComponents.length && !getComponents.isLoading && t("No components found with active filters")}
```
lencodes, line 142: The table headers shouldn't be visible at this point:

image

lencodes (changes requested):
Very nice!

pwa/src/templates/components/ComponentsTemplate.tsx
```diff
@@ -1,20 +1,21 @@
...
+import { InputText, Container, SelectMultiple } from "@conduction/components";
+import { ComponentResultTemplate } from "../componentResult/ComponentResultsTemplate";
 
 interface ComponentsTemplateProps {
   defaultTypeFilter?: string;
 }
 
 export const ComponentsTemplate: React.FC<ComponentsTemplateProps> = ({ defaultTypeFilter }) => {
   const [components, setComponents] = React.useState<any[]>([]);
+  const [display, setDisplay] = React.useState<string>("table");
```
lencodes, line 18: We can use the created type here instead of `<string>`.

PR #25, opened 2022-07-11
Search Input in landing page

hghanbari:


lencodes (changes requested):
Looks good! Left some comments.

Also: as discussed in our call, we want the new layout for the homepage.

PR #27, opened 2022-07-12
feature/OC-2/results-multi-display

remko48:
Added layer to multi display.

Note: because the data doesn't come with a `layer` the Layers tab shows all the components at every layer.


pwa/src/services/stringToCssModuleSelector.tsx
```diff
@@ -0,0 +1,11 @@
...
+  function capitalize(str: any) {
+    return str.charAt(0).toUpperCase() + str.slice(1);
+  }
+  function unCapitalize(str: any) {
+    return str.charAt(0).toLowerCase() + str.slice(1);
+  }
+
+  const newString = string.split(" ").map(capitalize).join(" ").replace(/\s/g, "").split(" ").map(unCapitalize);
+  return newString;
+};
```
lencodes, line 11: This seems quite complex. We have `lodash` included in our project, and I think [this function](https://www.geeksforgeeks.org/lodash-_-camelcase-method/) within it does the same.

pwa/src/templates/components/ComponentsTemplate.module.css
```diff
@@ -25,6 +25,15 @@
   align-items: center;
 }
 
+.resultsDisplaySwitchButtons * {
```
lencodes, line 28: You could make this rule more specific by adding a `>`.

pwa/src/templates/components/ComponentsTemplate.tsx
```diff
@@ -88,9 +103,16 @@ export const ComponentsTemplate: React.FC = () => {
         </FormField>
       </form>
 
-      {filteredComponents.length > 0 && <ComponentResultTemplate results={filteredComponents} type={display} />}
+      {getComponents.isSuccess && (
+        <>
+          {filteredComponents.length > 0 && (
```
lencodes, line 108: `filteredComponents.length` can only be `> 0` if `getComponents.isSuccess`. Therefore, we've created a redundant check here.

A possible solution would be:

```
{filteredComponents.length > 0 && (
        <ComponentResultTemplate components={filteredComponents} type={filters.resultDisplayLayout} />
      )}

      {!filteredComponents.length && !getComponents.isLoading && t("No components found with active filters")}
      {getComponents.isLoading && <Skeleton height="200px" />}
```

pwa/src/templates/templateParts/resultsTemplates/layers/LayersResultTemplate.tsx
```diff
@@ -0,0 +1,62 @@
...
+  return (
+    <QuoteWrapper borderWidth="20px">
+      <Heading1>{title}</Heading1>
+
+      <div className={styles.Container}>
+        {components.map((component) => (
+          <div
+            key={component.id}
+            onClick={() => navigate(`/components/${component.id}`)}
+            className={clsx(styles.Components, styles[stringToCssModuleSelector(component.developmentStatus)])}
```
lencodes, line 54: Tested this with:

```
...
import _ from lodash;

... 

className={clsx(..., styles[_.camelCase(component.developmentStatus])}
```

Seems to work 👍🏼

lencodes: That would make the added service redundant.

pwa/src/templates/templateParts/resultsTemplates/layers/LayersResultTemplate.tsx
```diff
@@ -0,0 +1,62 @@
...
+    setInteractieComponents(components);
+    setProcesComponents(components);
+    setIntegratieComponents(components);
+    setServicesComponents(components);
+    setDataComponents(components);
+  }, [components]);
+
+  return (
+    <>
+      {interactieComponents.length !== 0 && <Layer title="Interactie" components={interactieComponents} />}
```
lencodes, line 30: For consistency, I would change this operator from `!==` to `>`.

pwa/src/templates/templateParts/resultsTemplates/layers/LayersResultTemplate.tsx
```diff
@@ -0,0 +1,62 @@
...
+      <Heading1>{title}</Heading1>
+
+      <div className={styles.Container}>
+        {components.map((component) => (
+          <div
+            key={component.id}
+            onClick={() => navigate(`/components/${component.id}`)}
+            className={clsx(styles.Components, styles[stringToCssModuleSelector(component.developmentStatus)])}
+          >
+            <Link>{component.name}</Link>
```
lencodes, line 56: Since the entire card already has an onClick, I think wrapping this in a Link might be redundant. 

Also, the card can get a :hover { cursor: pointer }; so that there's visual feedback when hovering.

pwa/src/templates/templateParts/resultsTemplates/layers/LayersResultTemplate.module.css
```diff
@@ -0,0 +1,41 @@
+.Container {
```
lencodes, line 1: All classNames may start with a lowercase letter.

pwa/src/templates/templateParts/resultsTemplates/layers/LayersResultTemplate.module.css
```diff
@@ -0,0 +1,41 @@
...
+.Container > * {
+  margin-block-end: var(--nlportal-space-block-lg);
+}
+
+.Container > *:not(:last-child) {
+  margin-inline-end: var(--nlportal-space-block-xl);
+}
+
+.Components {
+  display: inline-block;
```
lencodes, line 15: I don't think adding `inline-block` does anything here.

pwa/src/templates/templateParts/resultsTemplates/layers/LayersResultTemplate.module.css
```diff
@@ -0,0 +1,41 @@
...
+  background-color: rgb(245, 245, 245);
+  box-shadow: rgb(33 33 33 / 25%) 0px -3px 0px inset;
+  padding: 14px;
+}
+
+.Components > a {
+  text-decoration: none;
+}
```
lencodes, line 23: If we decide to remove the `Link` element around the name, these rules would become redundant.

pwa/src/templates/templateParts/resultsTemplates/layers/LayersResultTemplate.tsx
lencodes, line 54: This is a singular `component` and therefore not `.componentS`, I think.

pwa/src/templates/templateParts/resultsTemplates/layers/LayersResultTemplate.module.css
```diff
@@ -0,0 +1,41 @@
...
+  border-right-width: 8px;
+  border-right-style: solid;
+  border-right-color: var(--conduction-color-gepland);
+}
+
+.gewenst {
+  border-right-width: 8px;
+  border-right-style: solid;
+  border-right-color: var(--conduction-color-gewenst);
+}
```
lencodes, line 41: I'd opt to combine these `border-right-width` and `border-right-style` rules in one block. Right now, if we would want to change the style or width, we have to update all these rules.

(I tested just adding it to `.component`, however, that created a black border on the unselected)

A possible solution would be;

```
.beta,
.gewenst,
.gepland {
  border-right-width: 8px;
  border-right-style: solid;
}
```

pwa/src/templates/templateParts/resultsTemplates/layers/LayersResultTemplate.module.css
lencodes: There's quite some hardcoded css values here. Since this component is not yet refined, I think having some hardcoded values is fine (as long as we are aware of them).

However, I do suggest creating some local variables, like so:

```
:root {
  --layer-component-padding: 14px;
  ... 
}
```

lencodes: Makes it easier for us to, when the time's there, extract this component into the component package.

PR #29, opened 2022-07-12
gegeven Conduction en Rotterdam

hghanbari:


lencodes (changes requested):
Nice, did leave some comments.

lencodes (changes requested):
You didn't resolve all comments.

pwa/src/templates/templateParts/footer/FooterTemplate.tsx
```diff
@@ -21,7 +22,8 @@ export const FooterTemplate: React.FC = () => {
         </div>
 
         <div className={styles.imagesContainer}>
-          <img src={LogoRotterdam} />
+          <img onClick={() => navigate("https://www.rotterdam.nl/")} src={LogoRotterdam} />
+          <img onClick={() => navigate("https://www.conduction.nl/")} src={LogoConduction} />
```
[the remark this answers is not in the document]
lencodes: This may open in a new tab, instead of the current.

pwa/src/templates/templateParts/footer/FooterTemplate.module.css
```diff
@@ -14,6 +14,7 @@
 
 .imagesContainer {
   display: flex;
+  justify-content: space-evenly;
```
lencodes, line 17: I think the images may be aligned left (with some margin between them).

lencodes (changes requested):
When hovering an image, please add a :hover staat with `cursor: pointer`.

pwa/src/templates/templateParts/footer/FooterTemplate.module.css
hghanbari: Ik maak 
.imagesContainer > :not(:last-child) {
  margin-right: var(--nlportal-space-inline-xl);
}

pwa/src/templates/templateParts/footer/FooterTemplate.tsx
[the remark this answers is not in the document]
hghanbari: ik maak (window.open)

pwa/src/templates/templateParts/footer/FooterTemplate.module.css
```diff
@@ -14,6 +14,11 @@
 
 .imagesContainer {
   display: flex;
+  cursor: pointer;
```
lencodes, line 17: Because you've set this rule to the entire `imagesContainer`, the cursor becomes a pointer anywhere when hovering anywhere over the container. This means that the user gets the visual feedback of a changing cursor when there's nothing yet clickable.

To fix this, we can set it to only the logo's. Also, we can only set it on the `:hover` event, like so:

```
.imagesContainer > *:hover {
  cursor: pointer;
}
```

lencodes (approved):
Nice @hghanbari.

**I did leave one more comment, please to resolve that.**

Otherwise, approved 👍🏼

PR #32, opened 2022-07-14
Installaties added in tebel en kaart

hghanbari:


pwa/src/templates/templateParts/resultsTemplates/cards/CardsResultTemplate.tsx
```diff
@@ -21,8 +21,9 @@ export const CardsResultTemplate: React.FC<CardsResultTemplateProps> = ({ compon
           labelsWithIcon={[
             { label: "Laag", icon: <HamburgerIcon /> },
             { label: "Organisatie", icon: <HouseIcon /> },
+            { label: "Installations", icon: <SettingsIcon /> },
```
lencodes, line 24: Every label is in Dutch here. So this should be "Installaties".

pwa/src/templates/templateParts/resultsTemplates/cards/CardsResultTemplate.tsx
```diff
@@ -21,8 +21,9 @@ export const CardsResultTemplate: React.FC<CardsResultTemplateProps> = ({ compon
...
           ]}
-          tags={[component.developmentStatus, component.softwareType]}
+          tags={[component.developmentStatus, component.softwareType , component.usedBy.length]}
```
lencodes, line 26: `usedBy` can be `null`.

In the used test data set, `usedBy` always contains an array. Therefore, this works now. However when we do get other data and it is set to `null`, this would break.

Because: `null.length` doesn't exist. 

To fix this: add an optional operator, like so: `component.usedBy?.length`.

Also, once we no longer can get an error, we still want to make sure the frontend receives a value to show (if `null` the value to the tag should be `0`).

Maybe a final solution could be:

```
component.usedBy ? component.usedBy.length : 0
```

pwa/src/templates/templateParts/resultsTemplates/table/TableResultTemplate.tsx
```diff
@@ -45,6 +47,10 @@ export const TableResultTemplate: React.FC<LayersResultTemplateProps> = ({ compo
               <Tag tag="Organisatie" />
             </TableCell>
 
+            <TableCell>
+              <Tag tag={component.usedBy.length}  />
```
lencodes, line 51: Same problem here. `usedBy` can be `null` and this would break.

pwa/src/templates/templateParts/resultsTemplates/cards/CardsResultTemplate.tsx
lencodes: You could use the translation string here as well.

lencodes (changes requested):
Nice @hghanbari, did have some comments.

lencodes (approved):
Nice @hghanbari 👍🏼

PR #35, opened 2022-07-14
error style tabel gefixd 

hghanbari:


pwa/src/templates/templateParts/searchComponent/SearchComponentTemplate.tsx
```diff
@@ -40,7 +40,6 @@ export const SearchComponentTemplate: React.FC = () => {
           {t("Search")}
         </Button>
         <Button
-          className={styles.submit}
```
lencodes, line 43: This seems like an unintentional change.

lencodes (approved):
Nice. 

@hghanbari I left one comment about a change I do not think should not be part of this pull request. Please check that before merging.

pwa/src/templates/templateParts/searchComponent/SearchComponentTemplate.tsx
hghanbari: Because it was among console errors

PR #61, opened 2022-08-05
OC-142/API_page

remko48:


pwa/src/pages/apiDocs.tsx
```diff
@@ -0,0 +1,9 @@
+import { PageProps } from "gatsby";
```
lencodes, line 1: I don't think you actually need those.

pwa/src/templates/apiDocs/ApiDocsTemplate.tsx
```diff
@@ -0,0 +1,32 @@
...
+          dolor lectus quis orci.
+          <span
+            className={styles.link}
+            onClick={() => {
+              navigate("#");
+            }}
+          >
+            <Link icon={<ExternalLinkIcon />} iconAlign="start">
+              Lees meer.
+            </Link>
```
lencodes, line 26: I'd add this link somewhere in between the `LeadParagraph`, so it becomes part of the content instead of it being just after everything.

Also, the label can be "bekijk de ReDoc".

pwa/src/templates/apiDocs/ApiDocsTemplate.tsx
```diff
@@ -0,0 +1,32 @@
...
+import { Container } from "@conduction/components";
+import { Heading1, LeadParagraph, Link } from "@gemeente-denhaag/components-react";
+import { navigate } from "gatsby";
+import { ExternalLinkIcon } from "@gemeente-denhaag/icons";
+
+export const ApiDocsTemplate: React.FC = () => {
+  return (
+    <Container layoutClassName={styles.container}>
+      <section className={styles.section}>
+        <Heading1>Api</Heading1>
```
lencodes, line 12: This label can become "API Documentatie"

pwa/src/templates/templateParts/footer/FooterTemplate.tsx
```diff
@@ -97,7 +97,7 @@ export const FooterTemplate: React.FC<FooterTemplateProps> = ({ layoutClassName
 
             <li
               onClick={() => {
-                navigate("#");
+                navigate("/apiDocs");
```
lencodes, line 100: Routes are usually snake cased, not camel.

So in this case that'd become `api-docs` instead of `apiDocs`.

Besides that, I think it would be better to create a subroute `documentation`.

So this page could land on `/documentation/api`.

lencodes (changes requested):
Nice start, did leave some comments.

I think the most important part would be the routing, you can create a folder `documentation` to create the updated routing.

lencodes (approved):

PR #62, opened 2022-08-05
OC-143/Usage-Page

remko48:


lencodes (changes requested):
Looks good @remko48.

I have one final comment: I'd stay way from using so many `br` elements. I think we could just create another `Paragraph` everywhere that you're now adding `<br><br><br><br>`.

Afterwards, I think we can merge this.

PR #64, opened 2022-08-09
Rewrite the quick start readme, and add some information about Helm

rjzondervan:
This in order to be a draft for the content of the installation page (although that will have to be translated to Dutch)

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,15 +25,40 @@ $ npm run develop
 ```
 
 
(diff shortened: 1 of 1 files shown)


lencodes (approved):
Looks good to me @rjzondervan.

We will do some translations when adding it to the frontend and add some links where needed (e.g. we're referring to the "Use this template" button which only makes sense in the context of reading this documentation on Github).

PR #65, opened 2022-08-10
feature/OC-45/rating-pie-chart

remko48:


pwa/src/templates/templateParts/ratingIndicator/RatingIndicatorTemplate.tsx
```diff
@@ -0,0 +1,56 @@
...
+  });
+
+  const calculatePercentage = (rating: number, highestPossibleRating: number) => {
+    const percentage = (100 * rating) / highestPossibleRating;
+    return percentage;
+  };
+
+  const ratingColor = getComputedStyle(document.documentElement).getPropertyValue(
+    "--web-app-color-background-page-header",
+  );
```
lencodes, line 33: This is very nicely found, and I don't see another way for us to retrieve our design tokens. However, because we'll be having to extract design-token-values within our javascript more often, I think a service might be a nice addition.

In the `services` folder we could add a new service with (basically doing the exact same thing you've already built here, but making it a bit more generic):

```
export const getTokenValue = (tokenKey: string) => {
  return getComputedStyle(document.documentElement).getPropertyValue(tokenKey);
};
```

Afterwards, we can add a `RatingIndicatorTemplate.modulde.css` with the following export (will be multiple exports, I just did this for one value to test):

```
:export {
  ratingColor: --web-app-color-background-page-header;
}
```

Within our Typescript, we can now retrieve this key (`styles.ratingColor`), and retrieve the value of this key, like so: 

```
labelStyle={{
    fontSize: "25px",
    fontFamily: ratingFont,
    fill: getTokenValue(styles.ratingColor), <== Using our new service and retrieving the key from css
}}
```

I think this will be a more robust (and reusable) way to retrieve the values.

pwa/src/templates/templateParts/ratingIndicator/RatingIndicatorTemplate.tsx
```diff
@@ -0,0 +1,56 @@
...
+  const ratingFont = getComputedStyle(document.documentElement).getPropertyValue("--web-app-primary-font-family");
+
+  return (
+    <PieChart
+      data={[{ value: 1, key: 1, color: ratingColor, title: `${rating}/${highestPossibleRating}` }]}
+      reveal={calculatePercentage(rating, highestPossibleRating)}
+      lineWidth={20}
+      background="#bfbfbf"
```
lencodes, line 41: This can become a design token.

pwa/src/templates/templateParts/ratingIndicator/RatingIndicatorTemplate.tsx
```diff
@@ -0,0 +1,56 @@
...
+      startAngle={270}
+      lengthAngle={360}
+      rounded
+      animate
+      animationDuration={1750}
+      label={({ dataEntry }) => dataEntry.title}
+      labelStyle={{
+        fontSize: "25px",
+        fontFamily: ratingFont,
+        fill: ratingColor,
```
lencodes, line 51: All of these can be design-tokens.

lencodes (changes requested):
Nice @remko48.

I left some inline comments.

Also, I think we should update the layout of the rating to, the link may link to the `/about` page: 

image

lencodes (approved):
Nice! Approved.

PR #68, opened 2022-08-17
feature/OC-14/UPL-display-on-detailspage

remko48:


pwa/src/templates/componentDetail/ComponentsDetailTemplate.tsx
```diff
@@ -96,8 +96,22 @@ export const ComponentsDetailTemplate: React.FC<ComponentsDetailTemplateProps> =
                   <TableCell>Rhoncus aenean vel elit scelerisque mauris</TableCell>
                 </TableRow>
                 <TableRow>
-                  <TableHeader>UPL</TableHeader>
-                  <TableCell>Sit amet nisl suscipit adipiscing bibendum est ultricies</TableCell>
+                  <TableHeader>{t("Products")}</TableHeader>
+                  <TableCell>
+                    {_getComponent.data.embedded.nl.upl.map((product: string, idx: any) => (
```
lencodes, line 101: 1. It's possible for `upl` to be unset. Might not be the case within the test data, but it could happen in the real data set. Therefore we cannot `.map()` on it without doing a check (mapping on `null` will break).

If `.embedded.nl.` is always available (which I think it should be, even though their values might be nullable, the path should be there)l. We could simply add an optional check, like so: `.nl?.upl.map( ... )`.

2. `idx` is of type `number`.

pwa/src/templates/componentDetail/ComponentsDetailTemplate.tsx
```diff
@@ -96,8 +96,22 @@ export const ComponentsDetailTemplate: React.FC<ComponentsDetailTemplateProps> =
...
+                      <a
+                        key={idx}
+                        href="http://standaarden.overheid.nl/owms/terms/AangifteVertrekBuitenland"
+                        target={"_blank"}
+                      >
```
lencodes, line 106: Even though this works (nothing wrong with it either), we've been doing outbound links in another way within the project. For consistency I'd recommend updating this to:

```
<span onClick={() => open("url here")} > // NOTE: This is the span element you already have!
  <Link icon={<ExternalLinkIcon />} iconAlign="start">
    {product},
  </Link>
</span>
```

pwa/src/templates/componentDetail/ComponentsDetailTemplate.tsx
```diff
@@ -96,8 +96,22 @@ export const ComponentsDetailTemplate: React.FC<ComponentsDetailTemplateProps> =
...
                 <TableRow>
-                  <TableHeader>UPL</TableHeader>
-                  <TableCell>Sit amet nisl suscipit adipiscing bibendum est ultricies</TableCell>
+                  <TableHeader>{t("Products")}</TableHeader>
+                  <TableCell>
+                    {_getComponent.data.embedded.nl.upl.map((product: string, idx: any) => (
+                      <a
+                        key={idx}
+                        href="http://standaarden.overheid.nl/owms/terms/AangifteVertrekBuitenland"
+                        target={"_blank"}
```
lencodes, line 105: Small detail: not relevant once you've updated it to underneath feedback, but: a string prop doesn't require the `{}` syntax, so: `target="_blank"` would work as well.

lencodes (changes requested):
Looks good @remko48. Left some comments.

PR #71, opened 2022-08-19
feature/GIT-189/card-component

remko48:


pwa/src/components/componentCard/ComponentCard.module.css
```diff
@@ -67,7 +67,7 @@
   --web-app-tag-background-color: #dd3c49;
   --web-app-tag-color: var(--web-app-color-white);
 }
-.procesCategory .processCategory {
+.procesCategory, .processCategory {
```
lencodes, line 70: For readability, I'd opt to put comma separated rules underneath each other, like so:

```
.process,
.other {
  ...
}
```

pwa/src/components/tag/Tag.tsx
```diff
@@ -21,8 +21,8 @@ interface TagIconLinkProps {
   href: string;
 }
 
-export const TagIconLink: React.FC<TagIconLinkProps> = ({ icon, href, tag }) => (
-  <div className={clsx(styles.tag, styles.tagIconLink)} onClick={() => navigate(href)}>
+export const TagIconLinkExternal: React.FC<TagIconLinkProps> = ({ icon, href, tag }) => (
+  <div className={clsx(styles.tag, styles.tagIconLink)} onClick={() => open(href)}>
     {tag && <span>{tag}</span>}
     <span>{icon}</span>
   </div>
```
lencodes, line 28: We're duplicating a lot of code this way (I understand why you did it though).

I think another possible solution would be to have the following props:

```
interface TagProps {
  content: JSX.Element; // icon? label? doesn't matter anymore, let the implementation decide
  onClick: () => any; // external? internal? doesn't matter anymore, let the implementation decide
}
```

pwa/src/templates/componentDetail/ComponentsDetailTemplate.tsx
```diff
@@ -94,42 +98,49 @@ export const ComponentsDetailTemplate: React.FC<ComponentsDetailTemplateProps> =
                         <Link icon={<ArrowRightIcon />} iconAlign="start">
                           {_getComponent.data.embedded?.legal.embedded?.repoOwner.name}
                         </Link>
+                        .
                       </span>
                     </div>
                   ) : (
-                    "Er is geen informatie beschikbaar"
+                    "Er is geen informatie beschikbaar."
```
lencodes, line 105: I'd opt out of using a ternary expression here.

Explicit expressions are easier to read, so in this case:

```
_getComponent.data.embedded?.legal.embedded?.repoOwner.name && <>...</>
!_getComponent.data.embedded?.legal.embedded?.repoOwner.name && <>...</>
```

lencodes: Also, for readability it might be worth it to save this within `state`. That way, we could have smaller variable names.

(dummy code)

```
const embedded = _getComponent.data.embedded;

// or even further

const legal = embedded.legal.embedded;
```

pwa/src/templates/componentDetail/ComponentsDetailTemplate.tsx
```diff
@@ -94,42 +98,49 @@ export const ComponentsDetailTemplate: React.FC<ComponentsDetailTemplateProps> =
...
-                <div>
-                  De broncode van dit component is te vinden op{" "}
-                  <span onClick={() => navigate("#")}>
-                    <Link icon={<ExternalLinkIcon />} iconAlign="start">
-                      GitHub
-                    </Link>
-                  </span>
-                </div>
+                <>
+                  {_getComponent.data.embedded?.url?.url ? (
```
lencodes, line 114: I'd opt out of the ternary expression.

lencodes (approved):
Looks good to me!

PR #79, opened 2022-08-25
feature/OC-188/component-and-organisation-detail-page-layer-accordion

remko48:
OC-188


lencodes (changes requested):
I forgot to hit the submit review button on this one.. 😓 

As discussed, the accordion logic (opening and closing and such) should be in one component and thereafter we can generate multiple accordions with the same component, just rendering different `children`.

pwa/src/components/badgeCounter/BadgeCounter.tsx
```diff
@@ -0,0 +1,16 @@
+import * as React from "react";
```
lencodes, line 1: This (much simpler) refactor is much better I think 👍🏼

pwa/src/components/badgeCounter/BadgeCounter.tsx
```diff
@@ -0,0 +1,16 @@
...
+import * as styles from "./BadgeCounter.module.css";
+import clsx from "clsx";
+
+interface BadgeCounterProps {
+  number: number | string;
```
lencodes, line 6: I think this component may just force what it wants as a prop, so either a `number` or a `string`, not both.

In this case, I'd opt for a `string`, since we're not doing any computations on it.

pwa/src/components/badgeCounter/BadgeCounter.module.css
```diff
@@ -0,0 +1,18 @@
...
+  font-size: var(--web-app-font-size-xs);
+  display: flex;
+  align-items: center;
+  justify-content: space-around;
+  background-color: var(--web-app-badge-counter-background-color);
+  color: var(--web-app-badge-counter-color);
+}
+.maxNumber {
+  font-size: var(--web-app-font-size-2xs);
+}
```
lencodes, line 18: I don't think `.maxNumber` is doing anything (it's not set anywhere; not should it. I think it's leftover from the previous implementation).

lencodes (changes requested):
Great refactor. 

Left some small comments. You can merge it afterwards 👍🏼

PR #80, opened 2022-08-26
feature/GIT-222/Component-detail-tags

remko48:


lencodes:
OC-181

pwa/src/templates/componentDetail/ComponentsDetailTemplate.module.css
```diff
@@ -20,12 +20,21 @@
   margin-block-end: var(--web-app-size-md) !important;
 }
 
-.tags {
+.layerTags {
```
lencodes, line 23: This does not contain only layerTags, maybe something along the lines of `layerAndCategoryContainer`? Working title 😄

pwa/src/templates/componentDetail/ComponentsDetailTemplate.tsx
```diff
@@ -25,22 +25,32 @@ import { GitHubLogo } from "../../assets/svgs/GitHub";
...
-  const [currentTab, setCurrentTab] = React.useState<number>(0);
   const { t } = useTranslation();
+  const [currentTab, setCurrentTab] = React.useState<number>(0);
 
   const queryClient = new QueryClient();
 
   const _useComponent = useComponent(queryClient);
   const _getComponent = _useComponent.getOne(componentId);
 
+  const layer: TCategories = t(_.upperFirst(_getComponent.data?.embedded?.nl.embedded.commonground.layerType));
```
lencodes, line 47: If we'd instead use `_.toLower()` here, could we get away with _not_ changing the `TCategories` keys to uppercase? Would be more consistent with our naming conventions.

pwa/src/templates/componentDetail/ComponentsDetailTemplate.tsx
```diff
@@ -61,22 +71,63 @@ export const ComponentsDetailTemplate: React.FC<ComponentsDetailTemplateProps> =
...
-                  {_getComponent.data.developmentStatus && (
-                    <Tag label={_.upperFirst(_getComponent.data.developmentStatus)} />
-                  )}
+                  <ToolTip tooltip="Laag">
+                    <Tag
+                      label={t(_.upperFirst(_getComponent.data.embedded?.nl.embedded.commonground.layerType))}
+                      icon={<FontAwesomeIcon icon={faLayerGroup} />}
+                    />
+                  </ToolTip>
                 </div>
```
lencodes, line 86: If we give the `Tooltip` component a `layoutClassName` prop, we could probably get away with not having a wrapper div here.

Also, right now we're setting the `tag-color` and `tag-background-color` within this wrapper div, why not just in the tag itself?

pwa/src/data/categories.tsx
```diff
@@ -26,10 +26,10 @@ import {
 } from "@fortawesome/free-solid-svg-icons";
 import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
 
-export type TCategories = "interactie" | "proces" | "integratie" | "services" | "data";
+export type TCategories = "Interactie" | "Proces" | "Integratie" | "Service" | "Data";
```
lencodes, line 29: I've left some other comments regarding this. But I think a general comment about this might be helpful.

I'd opt to keep this type lowercase (so: `"interactie", "..."`) instead of refactoring it to uppercase.

Afterwards, when we need it in another format (e.g. `upperFirst`) we can do so in the implementation.

This way, we stay more consistent with the way we create types (basically, this type holds a few keys and keys are not uppercase usually).

Note: it might be that the gateway returns its values with uppercase (I'm not sure), but those values we could `_.toLower()` before matching them with this type.

lencodes (changes requested):

PR #81, opened 2022-08-29
Development

remko48:


pwa/src/templates/organizationDetail/OrganizationDetailTemplate.tsx
```diff
@@ -33,9 +33,22 @@ export const OrganizationDetailTemplate: React.FC<OrganizationDetailTemplateProp
           <Heading1>Gemeente Rotterdam</Heading1>
 
           <LeadParagraph>
-            Morbi leo risus, porta ac consectetur ac, vestibulum at eros. Maecenas faucibus mollis interdum. Duis
-            mollis, est non commodo luctus, nisi erat porttitor ligula, eget lacinia odio sem nec elit. Vivamus sagittis
-            lacus vel augue laoreet rutrum faucibus dolor auctor.
+            <div className={styles.quote}>
```
lencodes, line 36: Couple of notes here:

- Do we (will we) actually get the data for a quote item from the backend? If not, this might work in this dummy-data scenario but it will not work with real data;
- Could we use our `<QuoteWrapper />` component here? Instead of making this custom element.

pwa/src/templates/organizationDetail/OrganizationDetailTemplate.tsx
```diff
@@ -33,9 +33,22 @@ export const OrganizationDetailTemplate: React.FC<OrganizationDetailTemplateProp
...
+              Een stad waar je ruimte ervaart. <br />
+              Bij een organisatie die je vrijheid en verantwoordelijkheid biedt. <br />
+              Voor werk waar je uitdaging voelt. <br />
+              Zodat jij en de stad blijven groeien. <br />
+              Voor en met alle Rotterdammers.
+            </div>
+            <br />
```
lencodes, line 43: Same goes for these `<br>` elements. Will we know where to place them? Will we receive `html` from the gateway?

pwa/src/templates/templateParts/filters/verticalFilters/VerticalFiltersTemplate.tsx
```diff
@@ -19,11 +19,13 @@ import {
   applicatiefuncties,
   referentieComponenten,
   organisations,
+  _categories,
```
lencodes, line 22: Could this be `categories`?

pwa/src/templates/templateParts/layerAccordion/LayerAccordionTemplate.module.css
```diff
@@ -9,3 +9,67 @@
...
+.dataCategory {
+  background-color: var(--web-app-layer-category-color-data) !important;
+  color: var(--web-app-color-black) !important;
+  border: none !important;
+}
+.dataCategory:hover {
+  background-color: var(--web-app-layer-category-color-data) !important;
+  color: var(--web-app-layer-color-data) !important;
+  cursor: pointer !important;
+}
```
lencodes, line 75: I think we should be able to override design tokens here, instead of using the `!important` statement everywhere.

Something like:

```
.processCategory {
  --web-app-current-color-rule: var(--web-app-layer-category-color-process);
  ...
}
```
Also, if we add a generic className to the different buttons, we can remove the duplicated rules (cursor and border) and add them to the generic className (since they're the same everywhere).

lencodes (changes requested):
Looks great @remko48, left some comments.

PR #83, opened 2022-08-30
feature/OC-183/Information-Blocks

remko48:


lencodes (changes requested):
Give me a quick call about this one.

(The code looks fine, I think we should re-think _how_ we want to present this data)

PR #85, opened 2022-08-31
feature/OC-182/styling-theme-switcher

remko48:


lencodes (changes requested):
I think a design like this (see screenshot) might work better (and would be less code). Also, the select element is available in our package (`SelectSingle`) so I don't think we have to duplicate that here.

Please do request a re-review after refactoring.

image

lencodes (approved):
Nice 👍🏼

PR #84, opened 2022-08-31
feature/OC-186/Organisation-card

remko48:

PR #88, opened 2022-09-08
feature/OC-192/OrganizationCard-implementation-CopmonentDetailPage

remko48:
Pull request #84  has been merged


lencodes:
Once #84 is merged, please re-assign me to this 👍🏼

pwa/src/data/organizations.ts
```diff
@@ -102,21 +102,21 @@ export const TEMPORARY_ORGANIZATIONS = [
...
+  //   owns: [],
+  //   uses: undefined,
+  //   token: "",
+  //   github: "",
+  //   gitlab: "",
+  //   website: "",
+  //   phone: "+31 10 267 16 25",
+  //   email: "info@unknown.unknown",
+  //   type: "unknown",
+  // },
```
lencodes, line 121: Unused code may be removed.

pwa/src/pages/components/[componentId]/ComponentsDetailPage.tsx
```diff
@@ -1,8 +1,22 @@
...
+
+  TempOrganization.push(sample);
+
+  return (
+    <>
+      {TempOrganization && (
+        <ComponentsDetailTemplate componentId={props.params.componentId} TempOrganization={TempOrganization} />
+      )}
+    </>
+  );
```
Comment (lencodes) [inline pwa/src/pages/components/[componentId]/ComponentsDetailPage.tsx:20]: In the `ComponentsDetailTemplate`, we always want to receive _one_ organisation (I think). Therefore, we never want it to be an array. I think there's some unwanted complexity here, and the following implementation also works:

```
const ComponentsDetailPage: React.FC<PageProps> = (props: PageProps) => {
  return (
    <ComponentsDetailTemplate componentId={props.params.componentId} organization={_.sample(TEMPORARY_ORGANIZATIONS)} />
  );
};
export default ComponentsDetailPage;
```

Note: I've changed the `TempOrganization` prop to `organization` because even though we're sending a temporary one, the `ComponentsDetailTemplate` doesn't have to know that.

Note: see the feedback on the `ComponentDetailTemplate`, because this new implementation will break some things there (which we'll fix in there).

pwa/src/templates/componentDetail/ComponentsDetailTemplate.module.css
```diff
@@ -52,7 +52,7 @@
 
 .cardsContainer {
   display: grid;
-  grid-template-columns: 2fr 2fr 2fr 1fr;
+  grid-template-columns: 2fr 1fr 1fr 1fr;
```
lencodes, line 55: It looks all right, however, a couple of notes (or thoughts):

1. We could remove the Github column and add the Github as an tag in the header (would seem more consistent);
2. The licenses would be better represented within a table, right now a tag within a column seems like an odd placement;

pwa/src/templates/componentDetail/ComponentsDetailTemplate.tsx
```diff
@@ -29,17 +29,22 @@ import _ from "lodash";
...
 import { ToolTip } from "../../components/toolTip/ToolTip";
 import { categories, TCategories } from "../../data/categories";
+import { categories as _categories } from "../../data/filters";
+import { OrganizationCard } from "../../components/organizationCard/OrganizationCard";
 import { GitHubLogo } from "../../assets/svgs/GitHub";
 import { BadgeCounter } from "../../components/badgeCounter/BadgeCounter";
 
 interface ComponentsDetailTemplateProps {
   componentId: string;
+  TempOrganization: any[];
```
lencodes, line 39: `organization: any;`

pwa/src/templates/componentDetail/ComponentsDetailTemplate.tsx
```diff
@@ -141,25 +146,22 @@ export const ComponentsDetailTemplate: React.FC<ComponentsDetailTemplateProps> =
...
+                components={{
+                  owned: organization.owns?.length.toString() ?? "0",
+                  supported: organization.supports?.length.toString() ?? "0",
+                  used: organization.uses?.length.toString() ?? "0",
+                }}
+                gitHub={organization.github}
+                gitLab={organization.gitlab}
+                type={organization.type}
+              />
+            ))}
```
lencodes, line 164: Right now, we're mapping over an array of organizations where we'll always want to receive one (I think..). If we refactor follow the feedback stated earlier in this review, we could completely remove the `.map()` here, and have the same outcome, like so:

```
<OrganizationCard
  title={{ label: organization.name, href: `/organizations/${organization.id}` }}
  description={organization.description}
  website={organization.website}
  logo={organization.logo}
  components={{
    owned: organization.owns?.length.toString() ?? "0",
    supported: organization.supports?.length.toString() ?? "0",
    used: organization.uses?.length.toString() ?? "0",
  }}
  gitHub={organization.github}
  gitLab={organization.gitlab}
  type={organization.type}
/>
```

lencodes (changes requested):
Left some comments 👍🏼 

I'm thinking that the columns header are filled with some content that might be better placed elsewhere (and the columns might be best to even be removed altogether), a very quick draft would come to something like (see other comments about it as well):

image

**Note:** if we feel like output is more important at this point than the overall UI/UX, feel free to ignore the column-related feedback.

PR #89, opened 2022-09-08
feature/OC-191/Implement-organisation-object

remko48:
Pull request #84  has been merged


lencodes:
I've unassigned myself, once #84 is merged, you can re-assign me on this 👍🏼

pwa/src/templates/organizationDetail/OrganizationDetailTemplate.tsx
```diff
@@ -36,127 +37,285 @@ export const OrganizationDetailTemplate: React.FC<OrganizationDetailTemplateProp
   const TempComponentsSupported = TEMPORARY_COMPONENTS.slice(8, 13);
   const TempComponentsUsed = TEMPORARY_COMPONENTS;
 
+  const tempOrganization = TEMPORARY_ORGANIZATIONS.find((organization) => {
+    return organization.id === organizationId;
+  });
+
```
lencodes, line 43: You could get rid of the explicit return statement here, like so:

```
const tempOrganization = TEMPORARY_ORGANIZATIONS.find((organization) => organization.id === organizationId);
```

pwa/src/templates/organizationDetail/OrganizationDetailTemplate.tsx
```diff
@@ -36,127 +37,285 @@ export const OrganizationDetailTemplate: React.FC<OrganizationDetailTemplateProp
...
-                </BadgeCounter>
-              }
-              value={2}
-            />
-          </Tabs>
-
-          <TabPanel className={styles.tabPanel} value="0">
-            <div className={styles.components}>
-              <ComponentCardsAccordionTemplate components={TempComponentsOwned} />
+      {tempOrganization && (
```
lencodes, line 46: Right now, we're duplicating the entire template based on wether or not the `tempOrganization` is set. If not, we're hardcoding the entire thing, if it is then we're using the dynamic one.

A better way to do this, is to make sure we have an organization set at all times. Right now, we're setting the `tempOrganization` like so:

`const tempOrganization = TEMPORARY_ORGANIZATIONS.find((organization) => organization.id === organizationId);`

Instead of thereafter directly returning the template (twice, depending on the state of `tempOrganization`, we could do the following:

```
if (!tempOrganization) {
 tempOrganization = TEMPORARY_ORGANIZATIONS[1]; // this is the Rotterdam one I think
}
```

Afterwards, the return statement of this component will **always** have an organization, so no need to check again _and_ no need to duplicate the template.

Note: I see that the current hardcoded part of this template has more data / layout (e.g. it wraps stuff in a `QuoteWrapper`, which wont happen with the test data. I'd still opt for above solution, if need be: update the temporary test data to mimic the intended layout.

pwa/src/templates/organizationDetail/OrganizationDetailTemplate.module.css
```diff
@@ -57,3 +76,9 @@
   margin-inline-start: var(--web-app-size-xs);
   align-items: inherit !important;
 }
+
+/* This styling is TEMPORARY and only for DEV & DEMO purposes */
+.oldLogo {
+  max-width: 100%;
+  max-height: var(--web-app-ornaziation-detail-old-logo-height);
+}
```
lencodes, line 84: This should become redundant, after other feedback is implemented.

pwa/src/templates/organizationDetail/OrganizationDetailTemplate.module.css
```diff
@@ -36,9 +36,28 @@
...
+
+.logoContainer {
+  display: flex;
+  align-items: center;
+  max-height: var(--web-app-ornaziation-detail-logo-height);
+  max-width: var(--web-app-ornaziation-detail-logo-width);
+}
+
 .logo {
+  vertical-align: middle;
```
lencodes, line 51: I'm unsure if you need this rule.

pwa/src/templates/organizationDetail/OrganizationDetailTemplate.module.css
```diff
@@ -36,9 +36,28 @@
   margin-block-end: var(--web-app-size-xs);
 }
 
+.headerContent {
+  flex: 2;
+}
+
+.logoContainer {
+  display: flex;
+  align-items: center;
```
lencodes, line 45: I'm unsure if you need these rules.

(Also: I would imagine that this is better outlined `flex-start`, which would be the same as not stating any rule at all).

pwa/static/themes/rotterdam.css
```diff
@@ -168,6 +168,12 @@
   --web-app-component-image-max-width: 300px;
   --web-app-component-rating-indicator-max-height: 160px;
 
+  --web-app-ornaziation-detail-logo-height: 300px;
+  --web-app-ornaziation-detail-logo-width: 350px;
+
+  /* This token is TEMPORARY and only for DEV & DEMO purposes */
+  --web-app-ornaziation-detail-old-logo-height: 300px;
```
lencodes, line 175: I think we should just remove this and keep to the dynamic template, instead of starting to add these kind of tokens.

If after reading all feedback, I'm missing something, and we do decide to keep that old implementation: I'd opt to _not_ create a design token for it (it's a temporary hardcoded thing, no need to make it dynamic).

Another note: we didn't add this to `theme.css.example`, which we should have (if we were to forget removing the functionality and token altogether, _and then_ create a new theme, this would break).

lencodes (changes requested):
Nice @remko48 

I left some comments.

PR #92, opened 2022-09-13
feature/OC-184/call-to-submit-component

remko48:


pwa/static/themes/rotterdam.css
```diff
@@ -44,6 +44,8 @@
   --web-app-size-3xl: var(--skeleton-size-3xl);
   --web-app-size-4xl: var(--skeleton-size-4xl);
 
+  --web-app-max-width-400: 400px;
```
lencodes, line 47: These tokens can exist, however, I think they should be more descriptive (as stated in other feedback), and I think they should be used within the _context_ and not within the component itself.

lencodes (changes requested):
Looks good @remko48, I did leave some comments.

Another note: I think that this placement (see screenshot) of the element might be better.

image

On another note: I don't think this happened on this pull request, however I noticed that the layer visualisation of component results broke:

image

pwa/src/templates/templateParts/submitComponent/SubmitComponentTemplate.module.css
```diff
@@ -0,0 +1,32 @@
...
+  font-weight: bold;
+  font-size: var(--web-app-font-size-xl);
+  margin-block-end: var(--web-app-size-xs);
+}
+
+.buttonContainer {
+  flex: 1;
+  align-self: center;
+  text-align: -webkit-right;
+}
```
lencodes, line 28: Whitespace

pwa/src/templates/templateParts/submitComponent/SubmitComponentTemplate.module.css
```diff
@@ -0,0 +1,32 @@
...
+.title {
+  font-weight: bold;
+  font-size: var(--web-app-font-size-xl);
+  margin-block-end: var(--web-app-size-xs);
+}
+
+.buttonContainer {
+  flex: 1;
+  align-self: center;
+  text-align: -webkit-right;
```
lencodes, line 27: We can do this without text-align or align-self. It should be positioning its children (the button), like so:

```
display: flex;
justify-content: flex-end;
align-items: center;
```

pwa/src/templates/templateParts/submitComponent/SubmitComponentTemplate.tsx
```diff
@@ -0,0 +1,35 @@
...
+export const SubmitComponentTemplate: React.FC<SubmitComponentTemplateProps> = ({ layoutClassName }) => {
+  return (
+    <div className={clsx(styles.container, layoutClassName && layoutClassName)}>
+      <div className={styles.header}>
+        <div className={styles.title}>Aan de slag met Open Catalogi</div>
+
+        <div>
+          Wilt u uw component op Open Catalogi aanbieden zodat andere uw component kunnen (her)gebruiken of bij dragen
+          aan de doorontwikkeling van uw component?
+        </div>
```
lencodes, line 21: I'd probably make this a `span` instead of a `div`, but doesn't matter too much.

I would however add some line-height to this element, because the text is very compressed.

I think `--web-app-line-height-sm` would be good 👍🏼

pwa/src/templates/templateParts/submitComponent/SubmitComponentTemplate.tsx
```diff
@@ -0,0 +1,35 @@
...
+      </div>
+      <div className={styles.buttonContainer}>
+        <Button
+          className={styles.button}
+          variant="secondary-action"
+          onClick={() => navigate("/documentation/usage")}
+          icon={<ArrowRightIcon />}
+        >
+          Component toevoegen!
```
lencodes, line 30: I'd remove the "!".

pwa/static/themes/rotterdam.css
```diff
@@ -178,6 +180,8 @@
   --web-app-relations-border: 1px solid #dddddd;
   --web-app-relations-background-color: rgb(251, 251, 251);
 
+  --web-app-feedback-description-max-width: 400px;
```
lencodes, line 183: This token doesn't seem to be implemented anywhere.

lencodes (approved):
Nice @remko48 👍🏼 

Left some small comments, don't think a re-review will be necessary afterwards.

PR #93, opened 2022-09-14
feature/OC-193/dependency-visualisation

remko48:


pwa/src/templates/templateParts/dependenciesTemplates/networkDependenciesTemplate/NetworkDependenciesTemplate.module.css
```diff
@@ -0,0 +1,24 @@
...
+  highlightLayerColorInteractie: --web-app-layer-category-color-interaction;
+  highlightLayerColorProces: --web-app-layer-category-color-process;
+  highlightLayerColorIntegratie: --web-app-layer-category-color-integration;
+  highlightLayerColorService: --web-app-layer-category-color-services;
+  highlightLayerColorData: --web-app-layer-category-color-data;
+}
+
+.networkContainer {
+  height: 500px;
+  width: var(--conduction-container-max-width);
```
lencodes, line 21: These two might be best as specific tokens for this element, e.g. `--web-app-relations-container-width` and height.

pwa/src/templates/templateParts/dependenciesTemplates/networkDependenciesTemplate/NetworkDependenciesTemplate.module.css
```diff
@@ -0,0 +1,24 @@
+:export {
```
lencodes, line 1: I'll just make one global comment about this: as stated earlier, I'd opt for another name for this (Relations) instead of Network. If we choose to adopt this change, please rename all the files as well.

pwa/src/templates/templateParts/dependenciesTemplates/networkDependenciesTemplate/NetworkDependenciesTemplate.module.css
```diff
@@ -0,0 +1,24 @@
...
+  highlightLayerColorIntegratie: --web-app-layer-category-color-integration;
+  highlightLayerColorService: --web-app-layer-category-color-services;
+  highlightLayerColorData: --web-app-layer-category-color-data;
+}
+
+.networkContainer {
+  height: 500px;
+  width: var(--conduction-container-max-width);
+  border: 1px solid #dddddd;
+  background-color: rgb(251, 251, 251);
```
lencodes, line 23: These may become design tokens.

pwa/src/templates/templateParts/dependenciesTemplates/networkDependenciesTemplate/NetworkDependenciesTemplate.module.css
```diff
@@ -0,0 +1,24 @@
...
+  layerColorInteractie: --web-app-layer-color-interaction;
+  layerColorProces: --web-app-layer-color-process;
+  layerColorIntegratie: --web-app-layer-color-integration;
+  layerColorService: --web-app-layer-color-services;
+  layerColorData: --web-app-layer-color-data;
+  borderLayerColorInteractie: --web-app-border-layer-color-interaction;
+  borderLayerColorProces: --web-app-border-layer-color-process;
+  borderLayerColorIntegratie: --web-app-border-layer-color-integration;
+  borderLayerColorService: --web-app-border-layer-color-services;
+  borderLayerColorData: --web-app-border-layer-color-data;
```
lencodes, line 11: I suggest to remove the borders altogether to (1) remove complexity in themes and (2) not introduce another color coupled to a layer which is _only_ visible in the relations view.

pwa/src/templates/templateParts/dependenciesTemplates/networkDependenciesTemplate/NetworkDependenciesTemplate.module.css
```diff
@@ -0,0 +1,24 @@
...
+  highlightLayerColorInteractie: --web-app-layer-category-color-interaction;
+  highlightLayerColorProces: --web-app-layer-category-color-process;
+  highlightLayerColorIntegratie: --web-app-layer-category-color-integration;
+  highlightLayerColorService: --web-app-layer-category-color-services;
+  highlightLayerColorData: --web-app-layer-category-color-data;
```
lencodes, line 16: I suggest to remove these highlight colors for the same reason as the borders. However, to keep a "highlighted" state, we could use opacity instead.

So, without a token, we could just set the opacity of the balls background. When highlighted.

lencodes: Note: I think it's even debatable to remove the highlighted state altogether since it doesn't do anything (you're not getting extra content or any other functionality out of it).

pwa/src/templates/templateParts/dependenciesTemplates/networkDependenciesTemplate/NetworkDependenciesTemplate.tsx
```diff
@@ -0,0 +1,122 @@
...
+            )
+          ],
+        ),
+        border: getTokenValue(
+          styles[
+            _.camelCase(`layerColor ${t(_.upperFirst(component.embedded?.nl?.embedded?.commonground?.layerType))}`)
+          ],
+        ),
+      },
+    },
```
lencodes, line 50: I would like to completely remove the highlight state and borders. To maintain simplicity in the frontend design _and_ to reduce complexity here.

pwa/src/templates/templateParts/dependenciesTemplates/networkDependenciesTemplate/NetworkDependenciesTemplate.tsx
```diff
@@ -0,0 +1,122 @@
...
+      from: component.id,
+      to: mainComponent.id,
+    };
+  });
+
+  const options = {
+    nodes: {
+      shape: "circle",
+      borderWidth: 2,
+      borderWidthSelected: 1,
```
lencodes, line 98: By setting this `borderWidth: 0` and removing the `...Selected`, the border is completely gone 👍🏼

pwa/src/templates/templateParts/dependenciesTemplates/networkDependenciesTemplate/NetworkDependenciesTemplate.tsx
```diff
@@ -0,0 +1,122 @@
...
+
+  const edges = nodes.map((component) => {
+    if (component.id === mainComponent.id) {
+      return {};
+    }
+    return {
+      from: component.id,
+      to: mainComponent.id,
+    };
+  });
```
lencodes, line 92: I'd go for this syntax.

```
const edges = nodes.map((component) => {
  if (component.id === mainComponent.id) return {};

  return {
    from: component.id,
    to: mainComponent.id,
  };
});
```

pwa/src/templates/templateParts/dependenciesTemplates/networkDependenciesTemplate/filters/NetworkDependenciesFiltersTemplate.tsx
```diff
@@ -0,0 +1,35 @@
+import * as React from "react";
```
lencodes, line 1: I'm wondering, we're building a complete new filters component here. However, we already have a filter component, the `LayerAccordionFiltersTemplate`.

Visually, they are an exact match (I think)?

The only difference would be their `handleClick` per item. 

Could this be as simple as to just rename the `LayerAccordionFiltersTemplate` to `LayerFiltersTemplate`, and implementing that one in both places?

lencodes: I see that in the current implementation (I think we've spoken about this, might be in an upcomming pull request) that we're using `badgeNumber` within the `LayerAccordionFiltersTemplate`, which would be the difference between these two.

I would suggest removing that `badgeNumber` altogether, and adding an optional `disabled` prop, which would make `LayerAccordionFiltersTemplate` less context-dependent _and_ make it implementable in the new context.

pwa/static/themes/rotterdam.css
```diff
@@ -144,6 +144,12 @@
   --web-app-layer-category-color-services: #d1e7dd;
   --web-app-layer-category-color-data: #e2d9f3;
 
+  --web-app-border-layer-color-interaction: #0047b3;
+  --web-app-border-layer-color-process: #981b25;
+  --web-app-border-layer-color-integration: #a6820c;
+  --web-app-border-layer-color-services: #3e755c;
+  --web-app-border-layer-color-data: #4b2b88;
+
```
lencodes, line 152: If we chose to remove the borders from the balls, these will become redundant.

lencodes (changes requested):
Looks awesome @remko48, very nice 😄 

I did have a few comments, see them below.

Also: if we want to demo this this afternoon, feel free to merge this (because it already works perfectly fine!) and afterwards implement this feedback!

PR #96, opened 2022-09-19
feature/OC-193/dependency-visualisation

remko48:


pwa/src/templates/templateParts/dependenciesTemplates/relationsDependenciesTemplate/RelationsDependenciesTemplate.tsx
```diff
@@ -1,12 +1,13 @@
 import * as React from "react";
```
lencodes, line 1: The layout sometimes "jumps", due to the frame re-rendering itself. Even though it keeps the same nodes. A possible solution would be to check if the components have changed _and_ if there's already a frame available _before_ re-rendering anything.

This idea of something that can only exist once is called a "Singleton", this might help (no idea if it will, didn't have time to really check it out): https://www.npmjs.com/package/react-singleton-hook.

pwa/src/templates/templateParts/dependenciesTemplates/relationsDependenciesTemplate/RelationsDependenciesTemplate.tsx
```diff
@@ -114,9 +93,16 @@ export const NetworkDependenciesTemplate: React.FC<LayersResultTemplateProps> =
...
+
   React.useEffect(() => {
-    container.current && new Network(container.current, { nodes, edges }, options);
-  }, [container, nodes, edges]);
+    const network = visJsRef.current && new Network(visJsRef.current, { nodes, edges }, options);
+    network?.on("doubleClick", (event: { nodes: string[] }) => {
+      if (event.nodes?.length === 1 && event.nodes[0] !== mainComponent.id) {
+        navigate(`/components/${event.nodes[0]}`);
+      }
+    });
```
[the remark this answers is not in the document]
lencodes: Note: the pop-up wouldn't be a perfect solution, but I think pretty low effort and high reward vs the current way it works.

lencodes (changes requested):
Nice changes @remko48!

Left some comments.

pwa/src/templates/templateParts/dependenciesTemplates/relationsDependenciesTemplate/RelationsDependenciesTemplate.tsx
```diff
@@ -93,16 +91,22 @@ export const RelationsDependenciesTemplate: React.FC<RelationsDependenciesTempla
...
+  const relationsContainerRef = React.useRef<HTMLDivElement>(null);
 
   React.useEffect(() => {
-    const network = visJsRef.current && new Network(visJsRef.current, { nodes, edges }, options);
+    const network =
+      relationsContainerRef.current && new Network(relationsContainerRef.current, { nodes, edges }, options);
+
+    if (!network) return;
+
     network?.on("doubleClick", (event: { nodes: string[] }) => {
```
lencodes, line 113: `network` exists here, no need to do the optional `?`

pwa/src/templates/templateParts/dependenciesTemplates/relationsDependenciesTemplate/RelationsDependenciesTemplate.tsx
```diff
@@ -15,38 +20,27 @@ interface LayersResultTemplateProps {
...
+export const RelationsDependenciesTemplate: React.FC<RelationsDependenciesTemplateProps> = ({
+  components,
+  mainComponent,
+}) => {
   const { t } = useTranslation();
 
-  const container = React.useRef(null);
+  const ToolTip = document.createElement("div");
+  const staticToolTipElement = renderToStaticMarkup(
+    <NodeToolTip tooltip="Dubbelklik het component om naar het component te gaan" />,
```
lencodes, line 31: Nice!

lencodes (approved):
Very nice!

I actually meant a pop-up card instead of the tooltip you've used. The Tooltip is a nice way of solving this as well, I think design-wise the pop-up card might look better.

Approved either way.

PR #97, opened 2022-09-19
feature/OC-194/header-only-on-homepage

remko48:


pwa/src/templates/templateParts/header/HeaderTemplate.tsx
```diff
@@ -12,6 +12,7 @@ import { ExternalLinkIcon } from "@gemeente-denhaag/icons";
 import clsx from "clsx";
```
lencodes, line 12: The code is fine like this.

The sub-pages _without_ the header look much better like this as well.

However, I do feel as if the header is very redundant at this point (a kind of redundant title with a kind of redundant subtitle, not adding much value).

What do you think of something like this:

image

lencodes (changes requested):

pwa/src/templates/templateParts/searchComponent/SearchComponentTemplate.tsx
```diff
@@ -23,7 +23,44 @@ export const SearchComponentTemplate: React.FC<SearchComponentTemplateProps> = (
...
+      "nl.commonground.layerType": [],
+      "nl.gemma.bedrijfsfuncties": [],
+      "nl.gemma.bedrijfsservices": [],
+      "nl.gemma.referentieComponenten": [],
+      "nl.gemma.applicatiefunctie": "",
+      "nl.upl": [],
+      "maintenance.type": "",
+      "legal.license": "",
+      "legal.mainCopyrightOwner": "",
+    });
```
lencodes, line 63: I think that the following does the same. A lot of properties in this object are optional, no need to set them with empty values 😄 

```
setFilters({
    resultDisplayLayout: filters.resultDisplayLayout,
    dependenciesDisplayLayout: filters.dependenciesDisplayLayout,
    currentPage: filters.currentPage,
} as IFilters);
```

lencodes: Note: you can do the same at the onSubmit, but then also add the search param, like so:

```
setFilters({
    search: data.name,
    resultDisplayLayout: filters.resultDisplayLayout,
    dependenciesDisplayLayout: filters.dependenciesDisplayLayout,
    currentPage: filters.currentPage,
} as IFilters);
```

lencodes (approved):
Very nice 😄 

I left two comments about resetting the filters. I don't think this will require a re-review afterwards.

PR #99, opened 2022-09-23
feature/OC-197/breadcrumbs

RalkeyOfficial:
Added breadcrumbs


pwa/src/templates/breadcrumb/BreadcrumbTemplate.tsx
```diff
@@ -0,0 +1,32 @@
...
+  const uppercaseCrumbs = crumbs.map((crumb: any) => ({
+    ...crumb,
+    crumbLabel: _.upperFirst(crumb.crumbLabel),
+  }));
+
+  const translatedBreadcrumbs = uppercaseCrumbs.map((crumb: any) => ({
+    ...crumb,
+    crumbLabel: t(crumb.crumbLabel),
+  }));
+
```
remko48, line 26: Nice job of translating the breadcrumbs! 
There is an easier way of doing this:
```
  const translatedBreadcrumbs= crumbs.map((crumb: any) => ({
    ...crumb,
    crumbLabel: t(_.upperFirst(crumb.crumbLabel)),
  }));
```
This way you can have both functions in one.

pwa/src/pages/documentation/index.tsx
```diff
@@ -0,0 +1,9 @@
+import { navigate } from "gatsby";
+import * as React from "react";
+
+const IndexPage = () => {
+  navigate("/");
+  return <></>;
+};
+
+export default IndexPage;
```
remko48, line 9: Nice catch!
I did notice that it still breaks on the organization's page breadcrumb. I suggest doing the same redirect there as you did here.

remko48 (changes requested):
Looks really good!

Left some small comments.

lencodes (changes requested):
I think it would be nice to have it look like this.

The background could be `--web-app-color-background-light`. Probably needs updates in component package.

image


Final note: I see that Documentatie > Over doesn't show "Documentatie" in the crumbs.

lencodes (approved):
Nice!

PR #105, opened 2022-09-28
feature/OC-203/nav-bar-active-status

RalkeyOfficial:
nog niet mergen totdat [deze pr](https://github.com/ConductionNL/conduction-components/pull/39) op conduction-components online is

### DIFF
--- a/pwa/src/styling/design-tokens/component-overrides.css
+++ b/pwa/src/styling/design-tokens/component-overrides.css
@@ -1,25 +1,15 @@
 :root {
-  --conduction-secondary-top-nav-item-padding: var(
-    --web-app-secondary-top-nav-item-padding
-  );
-  --conduction-secondary-top-nav-font-size: var(
-    --web-app-secondary-top-nav-font-size
-  );
+  --conduction-secondary-top-nav-item-padding: var(--web-app-secondary-top-nav-item-padding);
+  --conduction-secondary-top-nav-font-size: var(--web-app-secondary-top-nav-font-size);
   --conduction-secondary-top-nav-color: var(--web-app-secondary-top-nav-color);
-  --conduction-secondary-top-nav-font-weight: var(
-    --web-app-secondary-top-nav-font-weight
-  );
+  --conduction-secondary-top-nav-background-color-hover: var(--web-app-secondary-top-nav-background-color-hover);
+  --conduction-secondary-top-nav-font-weight: var(--web-app-secondary-top-nav-font-weight);
 
-  --conduction-primary-top-nav-item-padding: var(
-    --web-app-primary-top-nav-item-padding
-  );
+  --conduction-primary-top-nav-item-padding: var(--web-app-primary-top-nav-item-padding);
   --conduction-primary-top-nav-color: var(--web-app-primary-top-nav-color);
-  --conduction-primary-top-nav-background-color: var(
-    --web-app-primary-top-nav-background-color
-  );
-  --conduction-primary-top-nav-dropdown-border-radius: var(
-    --web-app-primary-top-nav-dropdown-border-radius
-  );
+  --conduction-primary-top-nav-background-color: var(--web-app-primary-top-nav-background-color);
+  --conduction-primary-top-nav-background-color-hover: var(--web-app-primary-top-nav-background-color-hover);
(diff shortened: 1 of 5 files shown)


lencodes (approved):
-

lencodes (changes requested):
This implementation is great 👍🏼.

Two notes before we can merge this though:

1. The changes within the component package itself still have some issues, review explains it all and
2. We didn't set the active and hover tokens within this project (I'd expect them within theme.example, and all themes).

pwa/static/themes/rotterdam.css
```diff
@@ -182,16 +182,12 @@
...
   --web-app-secondary-top-nav-font-size: var(--web-app-font-size-md);
   --web-app-secondary-top-nav-color: var(--web-app-color-black);
-  --web-app-secondary-top-nav-background-color-hover: auto;
   --web-app-secondary-top-nav-font-weight: var(--web-app-font-weight-light);
 
   /* @condcution: Primary Top Navigation */
   --web-app-primary-top-nav-item-padding: var(--web-app-size-md);
   --web-app-primary-top-nav-color: var(--web-app-color-white);
   --web-app-primary-top-nav-background-color: var(--rotterdam-color-green);
-  --web-app-primary-top-nav-background-color-hover: var(
```
lencodes, line 192: You've removed rules from `rotterdam.css` and from `utrecht.css`, whenever you do this, you should also remove them from `theme.css.example` as this is the template for all other implementation themes (e.g. rotterdam and utrecht, but could also be conduction unilever, whatever).

pwa/src/styling/design-tokens/component-overrides.css
```diff
@@ -6,9 +6,6 @@
     --web-app-secondary-top-nav-font-size
```
lencodes, line 6: I see you've removed the hover design tokens here, I'm assuming you did that because you added default values to the component package. Even though that _does_ work (right now) as intended, this is not the best way to do this.

**Component package**

- Includes generic components, with generic styling;
- This generic styling is completely customisable with design-tokens (`--conduction-....` tokens);
- **Important**: this initial, generic, styling is only _initial_ and not actually intended to be the final styling within an implementation (e.g. Open Catalogi).

**Implementation (e.g. Open Catalogi):**

- Whenever including a component from the component package (e.g. the `PrimaryTopNav`);
- We add all available tokens to `theme.css.example` (this is the example theme we can just copy and paste when making a new theme);
- We add new tokens from `theme.css.example` to all theme implementations (in this case: Utrecht and Rotterdam);
- We set the new values within `utrecht.css` and `rotterdam.css` to the values relative to that theme (they might be identical, but they don't _have_ to be, making it customisable);
- We add all new tokens within `component-overrides` to the relative `--conduction` token, to **override** the initial, generic, styling from the components.

The difference between what's stated above and what we've now done in this pull request: we're explicitly telling each theme how it should render the components, instead of implicitly just having the generic styling from the component package set the styling.

pwa/src/styling/design-tokens/component-overrides.css
```diff
@@ -6,9 +6,6 @@
...
   );
   --conduction-secondary-top-nav-color: var(--web-app-secondary-top-nav-color);
-  --conduction-secondary-top-nav-background-color-hover: var(
-    --web-app-secondary-top-nav-background-color-hover
-  );
```
lencodes, line 11: See the explanation comment about `.prettier` in your other pull request, the same issue is happening here.

lencodes (changes requested):

pwa/src/templates/templateParts/header/HeaderTemplate.tsx
```diff
@@ -22,69 +22,83 @@ interface HeaderTemplateProps {
 export const HeaderTemplate: React.FC<HeaderTemplateProps> = ({ layoutClassName }) => {
   const { t } = useTranslation();
   const [filters, setFilters] = React.useContext(FiltersContext);
+  const {
+    location: { pathname },
+  } = React.useContext(GatsbyContext);
```
lencodes, line 27: You've already declared `pathname` at the bottom of this file (I'd opt to keep this top one and remove the bottom one).

Due to already declaring it, this is breaking code. 

Note: whenever making a pull request, please check if the builds on Github have ran successfully, which is a good indicator for breaking code (see screenshot).

image

lencodes (changes requested):

PR #106, opened 2022-09-28
feature/OC-202/move-tooltip-component-to-component-package

RalkeyOfficial:
only merge once [this pr](https://github.com/ConductionNL/conduction-components/pull/40) is online


pwa/src/templates/templateParts/dependenciesTemplates/relationsDependenciesTemplate/RelationsDependenciesTemplate.tsx
```diff
@@ -6,10 +6,10 @@ import _ from "lodash";
 import { getTokenValue } from "../../../../services/getTokenValue";
 import { addNewLineToString } from "../../../../services/addNewLineToString";
 import { navigate } from "gatsby";
-import { NodeToolTip } from "../../../../components/toolTip/ToolTip";
 import { NotificationPopUp as _NotificationPopUp } from "@conduction/components";
 import "vis-network/styles/vis-network.css";
 import { renderToStaticMarkup } from "react-dom/server";
+import { NodeToolTip } from "@conduction/components/lib/components/toolTip/ToolTip";
```
lencodes, line 12: This is causing your build to break. See screenshot.

image

lencodes (changes requested):
Implementation looks good, however, your build did break.

In the future, please do double check if the Github build goes through.

Note: I think you've built a new ToolTip component, which you're trying to include in `RelationsDependenciesTemplate` but is not yet exported from the components package. Shouldn't all the other tooltips also become the new one?

PR #112, opened 2022-09-30
feature/OC-204/tooltip-gets-cut-off

RalkeyOfficial:


lencodes (changes requested):
I think we should have a quick call about the ToolTips, this pull request seems to be linked to the other two regarding ToolTips.

lencodes (changes requested):
I don't think we still need these changes. This feature branch has many conflicts and broken ToolTips. 

_If_ we want to update the tooltips within the project, we should use `react-tooltip`.

lencodes:
Closing this pull request due to the dated changes.

PR #113, opened 2022-09-30
feature/OC-212/prettier-update

remko48:


lencodes (approved):
Looks good to me.

One note: I don't see actual changes in the `.prettierrc` file. So I'm assuming it stayed the same but you've re-saved some files for consistency?

PR #114, opened 2022-10-03
feature/OC-213/Admin-login

remko48:


pwa/src/services/auth.tsx
```diff
@@ -11,15 +11,12 @@ export const isBrowser = (): boolean => typeof window !== "undefined";
 export const handleLogin = async (data: IUnvalidatedUser, API: APIService) => {
   if (!isBrowser()) return;
 
-  return await API.Login.login(data).then((res) => {
-    API.setAuthentication(res.data.jwtToken);
-    navigate("/");
-  });
+  API.setAuthentication("0b4c72dd-3530-4f34-bfde-c7cc3e27c328");
```
lencodes, line 14: Is this an actual key? I assume it's not.

Maybe for that reason we could call it something like `"TEMPORARY-DUMMY-KEY"`.

This way, it's abundantly clear that it's not some mistakenly hardcoded key.

pwa/src/templates/templateParts/header/HeaderTemplate.tsx
```diff
@@ -122,13 +132,22 @@ export const HeaderTemplate: React.FC<HeaderTemplateProps> = ({ layoutClassName
...
+        }
+      : {
+          label: "Login",
+          handleClick: () => {
+            navigate("/login");
+          },
+          icon: <FontAwesomeIcon icon={faCircleUser} />,
+        },
   ];
   const {
```
lencodes, line 152: I think this location variable should be set at the top of this file, for consistency.

pwa/src/templates/templateParts/header/HeaderTemplate.tsx
```diff
@@ -122,13 +132,22 @@ export const HeaderTemplate: React.FC<HeaderTemplateProps> = ({ layoutClassName
...
+          },
+          icon: <FontAwesomeIcon icon={faCircleUser} />,
+        }
+      : {
+          label: "Login",
+          handleClick: () => {
+            navigate("/login");
+          },
+          icon: <FontAwesomeIcon icon={faCircleUser} />,
+        },
```
lencodes, line 150: I'd suggest to remove this ternary expression, there are a couple of ways to do this, but maybe the following would be very usable:

```
const logoutMenuItem = { ... item comes here };
const loginMenuItem = { ... item comes here };

const secondaryTopNavItems = [ ... always available items here ];

if (isLoggedIn()) secondaryTopNavItems.push(logoutMenuItem);
if (!isLoggedIn()) secondaryTopNavItems.push(loginMenuItem);
```

pwa/src/templates/templateParts/header/HeaderTemplate.tsx
```diff
@@ -23,7 +24,7 @@ export const HeaderTemplate: React.FC<HeaderTemplateProps> = ({ layoutClassName
   const { t } = useTranslation();
   const [filters, setFilters] = React.useContext(FiltersContext);
 
-  const primaryTopNavItems = [
+  const PrimaryTopNavItems = [
```
lencodes, line 27: This should actually be `primaryTopNavItems`, I think.

lencodes (approved):
Left some comments. No need for a re-review.

PR #123, opened 2022-10-11
feature/OC-45/rating-indicator-show-rating

RalkeyOfficial:


pwa/src/templates/componentDetail/ComponentsDetailTemplate.tsx
```diff
@@ -195,13 +200,87 @@ export const ComponentsDetailTemplate: React.FC<ComponentsDetailTemplateProps> =
...
+              <div className={styles.overlay}>
+                <NotificationPopUp
+                  {...{ hide, isVisible }}
+                  title={"Rating"}
+                  description={
+                    <>
+                      {_getComponent.data.embedded?.rating?.rating ? (
+                        <p>{`${_getComponent.data.embedded?.rating?.rating}/${_getComponent.data.embedded?.rating?.maxRating}`}</p>
+                      ) : (
+                        <p>{t("No rating available")}</p>
```
lencodes, line 237: `p` tags are reserved to contain paragraphs of text (and even better used whenever in the context of e.g. an article with headings and such).

In this context however, we're really not displaying any type of content, therefore a `span` would be better suitable.

pwa/src/templates/componentDetail/ComponentsDetailTemplate.tsx
```diff
@@ -195,13 +200,87 @@ export const ComponentsDetailTemplate: React.FC<ComponentsDetailTemplateProps> =
...
+                      )}
+
+                      <div className={styles.popupDescription}>
+                        {_getComponent.data.embedded?.rating?.rating >= 1 && (
```
lencodes, line 241: You can check if something has a `.length`, if it does, it's an array with items, like so:

```
_getComponent.data.embedded?.rating.rating.length && ( ... )
```

pwa/src/templates/componentDetail/ComponentsDetailTemplate.tsx
```diff
@@ -195,13 +200,87 @@ export const ComponentsDetailTemplate: React.FC<ComponentsDetailTemplateProps> =
...
+                </>
               }
               layoutClassName={styles.infoCard}
             />
+            {isVisible && (
+              <div className={styles.overlay}>
+                <NotificationPopUp
+                  {...{ hide, isVisible }}
+                  title={"Rating"}
+                  description={
```
lencodes, line 232: Right now, we're building a RatingTemplate within the ComponentDetailTemplate. This file is now 580 lines, which is an indicator we should be separating some logic and templates.

I suggest creating a template part which holds this rating overview. And just adding this template part within this description.

lencodes:
This was merged in another PR.

PR #125, opened 2022-10-11
feature/OC-218/organisation-card-has-real-data

RalkeyOfficial:


pwa/src/templates/componentDetail/ComponentsDetailTemplate.tsx
```diff
@@ -173,21 +170,25 @@ export const ComponentsDetailTemplate: React.FC<ComponentsDetailTemplateProps> =
...
-                owned: organization.owns?.length.toString() ?? "0",
-                supported: organization.supports?.length.toString() ?? "0",
-                used: organization.uses?.length.toString() ?? "0",
-              }}
-              gitHub={organization.github}
-              gitLab={organization.gitlab}
-              type={organization.type}
-              layoutClassName={styles.organizationCardContainer}
-            />
+            {organizationData ? (
```
remko48, line 173: To get rid of this ternary expression we should use 
`{_getComponent?.data?.embedded?.url?.embedded?.organisation && (...)}` if it exists and 
`{!_getComponent?.data?.embedded?.url?.embedded?.organisation && (...)}` if it does not exist. 

We should do this for better readability and code consistency

pwa/src/templates/componentDetail/ComponentsDetailTemplate.tsx
```diff
@@ -173,21 +170,25 @@ export const ComponentsDetailTemplate: React.FC<ComponentsDetailTemplateProps> =
...
+                  supported: organizationData.supports?.length.toString() ?? "0",
+                  used: organizationData.uses?.length.toString() ?? "0",
+                }}
+                gitHub={organizationData.github}
+                gitLab={organizationData.gitlab}
+                type={organizationData.type}
+                layoutClassName={styles.organizationCardContainer}
+              />
+            ) : (
+              <div className={styles.organizationCardContainer} />
```
remko48, line 190: I think it would be better if we put a `<span>` in this empty space with "There is no organization available" in it to fill up the space and tell the user why there is so much space there

remko48 (changes requested):
Looks good left some small comments

PR #126, opened 2022-10-11
feature/OC-221/logo-has-real-image

RalkeyOfficial:


pwa/src/templates/componentDetail/ComponentsDetailTemplate.tsx
```diff
@@ -167,7 +167,9 @@ export const ComponentsDetailTemplate: React.FC<ComponentsDetailTemplateProps> =
             </div>
 
             <div className={styles.addToCatalogusContainer}>
-              <img src={grey} className={styles.componentImage} />
+              <div className={styles.logoContainer}>
+                <img src={_getComponent.data?.embedded?.url?.avatar_url ?? grey} className={styles.componentImage} />
```
lencodes, line 171: The default value (having a default value is a good idea) should have a more descriptive name, e.g. `componentPlacholderLogo`.

pwa/src/templates/componentDetail/ComponentsDetailTemplate.module.css
```diff
@@ -45,8 +45,16 @@
...
+  max-height: var(--web-app-component-detail-logo-height);
+  max-width: var(--web-app-component-detail-logo-width);
+  text-align: center;
+}
+
 .componentImage {
   max-width: var(--web-app-component-image-max-width);
+  max-height: 100%;
+  object-fit: contain;
 }
```
lencodes, line 58: Whenever setting the height and width of an image, using a container is a good idea. However, the container should be in control of the width and height of the image afterwards, which is not (really) the case here.

```
.logoContainer {
  width: var(--web-app-component-detail-logo-width); // not max-width
  height: var(--web-app-component-detail-logo-height); // not max-height
  // note: its children will always be 100% filling, so text-align is redundant.
}

.componentLogo {
  width: 100%; // always fill its container
  height: 100%; // always fill its container
  object-fit: cover;
}
```

Besides above refactoring, I'd suggest renaming `.componentImage` to either `.componentLogo` or just `.logo`. Since it's the child of the `logoContainer`.

lencodes: There is, however, the need to set a max-width on the `container`. I'd suggest setting this to a fixed value _without_ the use of a design token (it's a must-have fixed value, which should not be dynamically changed by the theme).

pwa/src/templates/componentDetail/ComponentsDetailTemplate.module.css
```diff
@@ -45,8 +45,16 @@
   flex: 2;
 }
 
+.logoContainer {
+  max-height: var(--web-app-component-detail-logo-height);
```
lencodes, line 49: (See other comment about how we can refactor this)

Note: when setting `max-height`, I'd expect the design token to reflect that (currently it's reflecting `height` and not `max-height`).

pwa/src/templates/componentDetail/ComponentsDetailTemplate.module.css
```diff
@@ -45,8 +45,16 @@
...
+  max-width: var(--web-app-component-detail-logo-width);
+  text-align: center;
```
lencodes, line 51: Whenever you use `text-align` on something that's not a text, it's probably best to look into (e.g.)

```
display: flex;
justify-content: center;
```

pwa/src/templates/organizationDetail/OrganizationDetailTemplate.module.css
```diff
@@ -41,8 +41,8 @@
 }
 
 .logoContainer {
-  max-height: var(--web-app-ornaziation-detail-logo-height);
-  max-width: var(--web-app-ornaziation-detail-logo-width);
+  max-height: var(--web-app-organization-detail-logo-height);
+  max-width: var(--web-app-organization-detail-logo-width);
 }
```
lencodes, line 46: Same feedback as the component logo and container apply here.

lencodes (changes requested):
Nice change @RalkeyOfficial, did have some comments.

pwa/src/templates/componentDetail/ComponentsDetailTemplate.module.css
```diff
@@ -45,8 +45,17 @@
...
 }
 
-.componentImage {
-  max-width: var(--web-app-component-image-max-width);
+.logoContainer {
+  height: 255px;
+  width: 300px;
+  display: flex;
+  justify-content: center;
+}
```
lencodes, line 53: I didn't necessarily mean that we should _completely_ hardcode these values. There is a solution which incorporates both fixed and dynamic values:

```
width: var(--dynamic-width);
height: var(--dynamic-height);
max-width: 300px; // fixed, like it is now
max-height: 255px; // fixed, like it is now
```

lencodes (approved):
For completion, I'd recommend refactoring the last comment.

Otherwise, approved 👍🏼

pwa/src/templates/componentDetail/ComponentsDetailTemplate.module.css
lencodes: This way, a theme could choose to make the logos smaller.

PR #127, opened 2022-10-12
feature/OC-219/Organization-API

remko48:


pwa/src/apiService/resources/organizations.ts
```diff
@@ -0,0 +1,59 @@
+import { Send } from "../apiService";
```
lencodes, line 1: A resource filename (and the classname, but that's already correct) should be singular.

pwa/src/apiService/resources/organizations.ts
```diff
@@ -0,0 +1,59 @@
...
+      value.forEach((value) => {
+        arrayParams += `&${key}[]=${value}`;
+      });
+
+      params += arrayParams;
+    }
+  }
+
+  return params;
+};
```
lencodes, line 59: This (seems to be) is a duplicate piece of logic which we're also using within the component resource. I see that the only difference is the filters we're deleting before setting the params.

I suggest creating a generic service to do this, to which you send the already _filtered_ filters object (so we do the deletions **before** passing them to the `filtersToQueryParams` function).

This way, the function can be generic, and _what_ we want to be added to the query params can be filtered by the implementing context.

pwa/src/templates/organizationDetail/OrganizationDetailTemplate.tsx
```diff
@@ -31,158 +33,174 @@ interface OrganizationDetailTemplateProps {
...
-              {tempOrganization.github && (
-                <ToolTip tooltip="GitHub">
-                  <Tag label={t("GitHub")} icon={<GitHubLogo />} onClick={() => open(tempOrganization.github)} />
-                </ToolTip>
-              )}
-              {tempOrganization.gitlab && (
-                <ToolTip tooltip="GitLab">
-                  <Tag label={t("GitLab")} icon={<GitLabLogo />} onClick={() => open(tempOrganization.gitlab)} />
-                </ToolTip>
+              {!_getOrganization.data.description && <LeadParagraph>{_getOrganization.data.description}</LeadParagraph>}
```
lencodes, line 48: I think this check shouldn't be inverted.

pwa/src/templates/organizationDetail/OrganizationDetailTemplate.tsx
```diff
@@ -31,158 +33,174 @@ interface OrganizationDetailTemplateProps {
...
-              {tempOrganization.phone && (
-                <ToolTip tooltip={"Telefoonnummer"}>
+            </div>
+
+            <div className={styles.headerOrganizationData}>
+              <div className={styles.logoContainer}>
+                {_getOrganization.data.logo && (
+                  <img className={styles.logo} src={_getOrganization.data.logo} alt="Organization logo" />
+                )}
+                {!_getOrganization.data.logo && <img src={grey} className={styles.logo} />}
```
lencodes, line 59: The name of `grey` could be more descriptive, e.g. `organizationPlaceholderImage`.

lencodes, line 59: Could this be simplified by doing: ``?

pwa/src/templates/organizationDetail/OrganizationDetailTemplate.tsx
```diff
@@ -31,158 +33,174 @@ interface OrganizationDetailTemplateProps {
...
-                <ToolTip tooltip={"EmailAddress"}>
+                <ToolTip tooltip={"ISO-27001"}>
                   <Tag
-                    label={tempOrganization.email}
-                    icon={<FontAwesomeIcon icon={faEnvelope} />}
-                    onClick={() => navigate(`mailto:${tempOrganization.email}`)}
+                    label={"ISO-27001"}
+                    icon={<FontAwesomeIcon icon={faCertificate} />}
+                    onClick={() => open("https://www.iso.org/isoiec-27001-information-security.html")}
                   />
```
lencodes, line 126: Are the licenses not coming back with the API yet? If not, could we create a issue for this?

lencodes (changes requested):
Nice changes! Did have some comments.

Whenever we cannot find an organisation, I think it'd be best to keep the UI a bit more consistent (since we've leaned into the "block" design), like so (note: I also changed the label):
image

I noticed we're still missing some of the actual implementations of the new API (unsure if we wanted to do that in this pull request or in another; if in another, please do create the issues):

1. Organisation filter dropdown;
2. Organisation components on the detail page, see screenshot

image

PR #133, opened 2022-10-14
feature/OC-230/add-detail-page

RalkeyOfficial:


remko48 (changes requested):
Looks a lot better but i have a couple of things.

from the previous pr the file ` /data/users.ts`  needs some prettifier.

PR #135, opened 2022-10-17
feature/OC-211/hergebruik-tab-has-organizationCard

RalkeyOfficial:


pwa/src/templates/componentDetail/ComponentsDetailTemplate.tsx
```diff
@@ -414,77 +417,28 @@ export const ComponentsDetailTemplate: React.FC<ComponentsDetailTemplateProps> =
...
+                      components={{
+                        owned: organization?.owns?.length.toString() ?? "0",
+                        supported: organization?.supports?.length.toString() ?? "0",
+                        used: organization?.uses?.length.toString() ?? "0",
+                      }}
+                      gitHub={organization?.github}
+                      gitLab={organization?.gitlab}
+                      type={organization?.type}
+                      layoutClassName={styles.organizationCardContainer}
+                    />
```
remko48, line 440: When showing the Organizations we we'll need to wrap it in a card grid like we do on the components page.
image

remko48 (changes requested):
Looks good and works but i did leave some comments.

PR #143, opened 2022-10-25
feature/OC-239/top-nav-refactor

RalkeyOfficial:


pwa/src/templates/templateParts/header/HeaderTemplate.tsx
```diff
@@ -116,6 +129,13 @@ export const HeaderTemplate: React.FC<HeaderTemplateProps> = ({ layoutClassName
             navigate("/documentation/standards");
           },
         },
+        {
+          label: t("Contact"),
+          current: pathname === "/documentation/contact",
+          handleClick: () => {
+            navigate("/documentation/contact");
```
remko48, line 136: Right now this navigate links to a non existing page. I think this meant to be a "#"

remko48 (changes requested):
Looks good just a small comment. 

This is out of scope for this issue but its probably a good idea to reset the filters everytime we click on one of the menu items. For example if click on `API's` within `Componenten` and then on `Initiatieven` everything works fine but when we go back from `Initiatieven` to `API's` no components can be found. This happens because another filter is active `Status`.

remko48 (approved):

PR #146, opened 2022-10-26
feature/OC-242/filters-get-reset

RalkeyOfficial:


lencodes (changes requested):
Even though this works as intended, I think we should refactor it to be _much_ less code and _more_ generic.

**First, we have a sort of base filters, which we always want to be set.**

These are set within the `filters.ts` context file, named `filters`. I'd opt to rename this, like so (we import these in some files, make sure to rename the imports as well!):

```
export const baseFilters = {
  resultDisplayLayout: "table",
  dependenciesDisplayLayout: "layer",
  landingDisplayLayout: "categories",
  currentPage: 1,
} as IFilters;
```

**Second, we want to _reset_ the filters on some clicks, and set only a specific filter.**

Instead of setting every key to some empty value within the `IFilters` interface, we could only set the `baseFilters` (they are required) and then set the specific filter that we want to activate, like so:

```
```

lencodes (approved):
Nice changes, much better maintainable.

I did spot a issue (never implemented feature) with pre-filling the fields. I'll create a issue on JIRA for this.

PR #161, opened 2022-11-01
feature/OC-277/components-is-mobile-friendly

RalkeyOfficial:


pwa/src/templates/components/ComponentsTemplate.module.css
```diff
@@ -32,12 +32,14 @@
...
 .filtersAndResultsContainer {
   display: flex;
+  flex-direction: column;
   gap: var(--web-app-size-lg);
 }
 
 .verticalFilters {
   flex: 1;
   min-width: 280px;
+  height: min-content;
```
lencodes, line 42: I don't think this is actually doing anything.

pwa/src/templates/components/ComponentsTemplate.module.css
```diff
@@ -32,12 +32,14 @@
 
 .filtersAndResultsContainer {
   display: flex;
+  flex-direction: column;
```
lencodes, line 35: I think this should be unset on desktop (or maybe on tablet already?)

pwa/src/templates/templateParts/filters/verticalFilters/VerticalFiltersTemplate.module.css
```diff
@@ -16,10 +16,29 @@
...
   font-size: var(--web-app-size-md);
+  cursor: pointer;
+}
+
+.toggleIcon {
+  font-size: var(--web-app-font-size-lg);
+  transition: transform 200ms ease-in-out;
+}
+
+.toggleIcon.active {
```
lencodes, line 35: Nicely done. A better name for the state would be `isActive`.

pwa/src/templates/templateParts/filters/verticalFilters/VerticalFiltersTemplate.module.css
```diff
@@ -16,10 +16,29 @@
   font-size: var(--web-app-size-md);
 }
 
+.trigger > *:not(:last-child) {
```
lencodes, line 19: The label and the icon are not aligned properly. An easy way to fix this would be:

```
.trigger {
  display: flex;
  align-items: center;
}
```

pwa/src/templates/templateParts/filters/verticalFilters/VerticalFiltersTemplate.module.css
```diff
@@ -16,10 +16,29 @@
...
+}
+
 .title {
   display: block;
   font-weight: var(--web-app-font-weight-bold);
   font-size: var(--web-app-size-md);
+  cursor: pointer;
+}
+
+.toggleIcon {
```
lencodes, line 30: This size is a bit too large for its label, I'd opt for a bit smaller icon.

pwa/src/templates/templateParts/filters/verticalFilters/VerticalFiltersTemplate.tsx
```diff
@@ -26,13 +26,22 @@ import {
...
+import { faChevronRight } from "@fortawesome/free-solid-svg-icons";
+import { GatsbyContext } from "../../../../context/gatsby";
 
 interface VerticalFiltersTemplateProps {
   layoutClassName?: string;
 }
 
 export const VerticalFiltersTemplate: React.FC<VerticalFiltersTemplateProps> = ({ layoutClassName }) => {
   const [filters, setFilters] = React.useContext(FiltersContext);
+  const [open, setOpen] = React.useState<boolean>(false);
```
lencodes, line 40: `isOpen, setIsOpen` would be a more descriptive (and easy to read) state name.

lencodes (changes requested):
Looks good @RalkeyOfficial 

Few inline comments, and this more global one:

I'd expect the layout to stay the same on dekstop (two columns, filters left, results right), but right now this is happening:
image

pwa/src/templates/components/ComponentsTemplate.module.css
RalkeyOfficial: this makes sure the grey background follows the filter collapsible when it collapses.
otherwise it would end up like this:
image

lencodes (changes requested):
You've fixed that the layout on desktop is right now, however, now the mobile version is broken. See screenshot.

Please do fully test changes before creating pull requests.

image

PR #198, opened 2023-01-16
Bugfix/oc 391/filters rework

lencodes:


pwa/src/templates/templateParts/filters/verticalFilters/VerticalFiltersTemplate.tsx
```diff
@@ -161,6 +153,11 @@ export const VerticalFiltersTemplate: React.FC<VerticalFiltersTemplateProps> = (
       maintenanceType: getSelectedItemFromFilters(maintenanceTypes, filters["maintenance.type"]),
       license: getSelectedItemFromFilters(licenses, filters["legal.license"]),
       organization: organizations && getSelectedItemFromFilters(organizations, filters["url.organisation.name"]),
+      interface: filters["nl.commonground.layerType"]?.includes("interface"),
+      process: filters["nl.commonground.layerType"]?.includes("process"),
+      integration: filters["nl.commonground.layerType"]?.includes("integration"),
+      service: filters["nl.commonground.layerType"]?.includes("service"),
+      data: filters["nl.commonground.layerType"]?.includes("data"),
```
lencodes, line 160: @remko48 

The bug regarding the layer filters originated here. When using the `reset` method, it resets the entire form with the values passed. However, when we don't pass values for specific fields, it will empty them.

A bit odd though, is that some fields kept working. Either way, it's fixed now :).

PR #204, opened 2023-01-19
feature/OC-428/catagories-details-results-solutions

RalkeyOfficial:


pwa/src/apiService/resources/applications.ts
```diff
@@ -30,4 +30,5 @@ const deletes = [
   { name: "resultDisplayLayout" },
   { name: "dependenciesDisplayLayout" },
   { name: "landingDisplayLayout" },
+  { name: "catagoryDisplayLayout" },
```
lencodes, line 33: This entire array is never used, so I think instead of adding this item to the array, we can just completely remove this array.

pwa/src/apiService/resources/components.ts
```diff
@@ -37,4 +37,5 @@ const deletes = [
   { name: "resultDisplayLayout" },
   { name: "dependenciesDisplayLayout" },
   { name: "landingDisplayLayout" },
+  { name: "catagoryDisplayLayout" },
```
lencodes, line 40: We're currently duplicating this code on four places. Even though this piece of logic is _very_ generic. It doesn't change within the different implementations (nor should it!).

Therefore, I suggest moving this elsewhere, like so (working code, dummy comments):

```
// filtersToQueryParams.ts

// First, we can simplify the current implementation of deleting keys (right now, we need an object to be passed, I think a simple string would do the trick)

Object.keys(filters)
    .filter((key) => filterKeysToRemove.includes(key))
    .forEach((key) => {
      delete filters[key];
    });

// .. other logic of the file

// List of keys that are in IFilters, but should never be sent to the gateway as query params <= KEEP THIS COMMENT :)
const filterKeysToRemove: string[] = [
  "resultDisplayLayout",
  "dependenciesDisplayLayout",
  "landingDisplayLayout",
  "catagoryDisplayLayout",
];
```

lencodes: This way, we only need to keep track of this logic in one file and we can add it a  simple string array.

pwa/src/components/componentCard/ComponentCard.module.css
```diff
@@ -88,7 +94,9 @@
 
 .servicesCategory,
 .serviceCategory {
-  --conduction-tag-background-color: var(--web-app-layer-category-color-services);
+  --conduction-tag-background-color: var(
+    --web-app-layer-category-color-services
+  );
```
lencodes, line 99: All these changes are prettier changes, which shouldn't be happening. Can you check that you're using the `.prettierrc` file from within this repository and then re-save this file?

pwa/src/components/componentCard/ComponentCard.module.css
```diff
@@ -98,3 +106,8 @@
 .dataCategory {
   --conduction-tag-background-color: var(--web-app-layer-category-color-data);
 }
+
+.unknownLayer {
+  --conduction-tag-background-color: var(--web-app-color-top-nav-grey);
```
lencodes, line 111: I don't think a `...-top-nav-grey` token should be used here. Two issues:

1. This is not a `top-nav`, so using a dedicated token for another component is not logical and can create issues later on (e.g. if the `top-nav` is later rebranded to be orange).
2. Giving such a specific name to a token, kind of beats te purpose of the token. Right now it _has_ to hold a `grey` value. What if someone wants to make a orange top navigation? The key should simply be `..color-top-nav`, and even more specific, it should be `-background-color-top-nav`.

pwa/src/context/filters.ts
```diff
@@ -3,11 +3,13 @@ import * as React from "react";
 export type TComponentResultsLayout = "table" | "cards" | "layer";
 export type TLandingDisplayLayout = "layer" | "categories";
```
lencodes, line 4: I think a more logical name for this "categories" view would be "cards", like the other ones. Because it are cards after all. And I don't think "categories" is a type of layout.

pwa/src/templates/categoriesDetailTemplate/CategoriesDetailTemplate.tsx
```diff
@@ -11,15 +11,17 @@ import { TEMPORARY_PORTFOLIOS } from "../../data/portfolio";
...
-import { faTags } from "@fortawesome/free-solid-svg-icons";
+import { faGripVertical, faLayerGroup, faTable, faTags } from "@fortawesome/free-solid-svg-icons";
+import { FiltersContext } from "../../context/filters";
+import { ComponentResultTemplate } from "../templateParts/resultsTemplates/ComponentResultsTemplate";
 
 interface CategoriesDetailTemplateProps {
   categoryId: string;
 }
 
 export const CategoriesDetailTemplate: React.FC<CategoriesDetailTemplateProps> = ({ categoryId }) => {
```
lencodes, line 23: This is a `CategoryDetailTemplate`*

pwa/src/templates/categoriesDetailTemplate/CategoriesDetailTemplate.tsx
```diff
@@ -11,15 +11,17 @@ import { TEMPORARY_PORTFOLIOS } from "../../data/portfolio";
...
+  const [filters, setFilters] = React.useContext(FiltersContext);
   const { t } = useTranslation();
 
   const portfolio = TEMPORARY_PORTFOLIOS.find((category) => {
```
lencodes, line 27: A category contains components, not really a portfolio I'd say.

_If_ this template was called `PortfolioDetailTemplate`, I'd expect this variable to be called `portfolioItems` perhaps. However, in this case, I'd probably opt for `components`.

pwa/src/templates/categoriesDetailTemplate/CategoriesDetailTemplate.tsx
```diff
@@ -30,12 +32,8 @@ export const CategoriesDetailTemplate: React.FC<CategoriesDetailTemplateProps> =
     TEMPORARY_DOMAINS.find((domain) => {
       return domain.title === portfolio.domain;
     });
-  const component =
-    portfolio &&
-    TEMPORARY_COMPONENTS.find((component) => {
-      return component.id === "aa78c93b-d57d-4d9a-b255-e8cfe26bde48";
-    });
-  const badgeNumber = 1;
+  const components = TEMPORARY_COMPONENTS;
```
lencodes, line 35: Shouldn't this just be the components that are filtered for the current category? See previous comment. 

If the test data doesn't allow it, just keep using `TEMPORARY_COMPONENTS` throughout the file. No need to do this one to one assignment.

pwa/src/templates/categoriesDetailTemplate/CategoriesDetailTemplate.tsx
```diff
@@ -30,12 +32,8 @@ export const CategoriesDetailTemplate: React.FC<CategoriesDetailTemplateProps> =
...
+  const badgeNumber = components.results.length;
```
lencodes, line 36: No need to assign this, I think, we could just do:

```
const badgeLabel = badgeNumber < 100 ? _.toString(components.results.length) : "99+";
```

pwa/src/templates/categoriesDetailTemplate/CategoriesDetailTemplate.tsx
```diff
@@ -30,12 +32,8 @@ export const CategoriesDetailTemplate: React.FC<CategoriesDetailTemplateProps> =
...
   const badgeLabel = badgeNumber < 100 ? _.toString(badgeNumber) : "99+";
   const maxItems = badgeNumber > 100;
```
lencodes, line 38: I think `maxItems` should already be set if it's 100, not only if it's more than 100.

So: `>= 100`.

pwa/src/templates/categoriesDetailTemplate/CategoriesDetailTemplate.tsx
lencodes: Edit: we could use the assignment later in the file, I think.

However, `badgeNumber` doesn't say much about _what_ it actually represents. Maybe `componentCount` instead?

pwa/src/templates/categoriesDetailTemplate/CategoriesDetailTemplate.tsx
lencodes: I'd opt to just remove this variable.

pwa/src/templates/categoriesDetailTemplate/CategoriesDetailTemplate.tsx
```diff
@@ -54,7 +52,35 @@ export const CategoriesDetailTemplate: React.FC<CategoriesDetailTemplateProps> =
...
+        >
+          <FontAwesomeIcon icon={faLayerGroup} />
+          {t("Layers")}
+        </Button>
+      </div>
+
+      {components && (
         <div className={styles.solutions}>
           <div className={styles.solutionsHeader}>
             <span className={maxItems && styles.maxNumber}>
```
lencodes, line 86: Instead of using `maxItems && ...`, I'd opt to use `componentCount >= 100 && ...`.

pwa/src/templates/categoriesDetailTemplate/CategoriesDetailTemplate.tsx
```diff
@@ -54,7 +52,35 @@ export const CategoriesDetailTemplate: React.FC<CategoriesDetailTemplateProps> =
...
+        </Button>
+        <Button
+          className={styles.buttonIcon}
+          variant={filters.catagoryDisplayLayout === "layer" ? "primary-action" : "secondary-action"}
+          onClick={() => setFilters({ ...filters, catagoryDisplayLayout: "layer" })}
+        >
+          <FontAwesomeIcon icon={faLayerGroup} />
+          {t("Layers")}
+        </Button>
+      </div>
```
lencodes, line 81: This might be best to be refactored into a separate component, because we're re-using this logic in some other places. Could you create a issue to do this?

Doesn't have to be in this pull request.

lencodes (changes requested):
Nice @RalkeyOfficial 

Left some comments.

pwa/src/templates/categoriesDetailTemplate/CategoriesDetailTemplate.tsx
RalkeyOfficial: a variable called `components` already exists inside this template
and calling it `components` does not make a whole lot of sense considering the data inside `TEMPORARY_PORTFOLIOS`

lencodes (approved):
Nice 👍🏼

PR #210, opened 2023-01-26
Feature/oc 451/programatically select radio

lencodes:
Rewrites logic of radio buttons in the filters. Was necessary due to radio buttons never being pre-selected when the filters were updated programatically. 

With this refactor, we use the power of react-hook-form more and decrease the need for custom code and logic.

(Note: we had a couple of `document.getElementById` queries in this logic, which is not really the _"React way"_ of doing things).


remko48 (approved):
That looks way better than my approach!

PR #214, opened 2023-01-30
feature/search-cleanup

remko48:
In this pr are two main features
    1. We make sure we have working data
    2. A hotfix for the components detail template crash on reading `0` of `undefined`

### DIFF
--- a/pwa/src/apiService/apiService.ts
+++ b/pwa/src/apiService/apiService.ts
@@ -40,19 +40,6 @@ export default class APIService {
     });
   }
 
-  public get searchClient(): AxiosInstance {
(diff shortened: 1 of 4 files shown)


lencodes (changes requested):
The changes look good.

I did however see the following issue: when clicking on a component in (e.g.) the `TableResultTable`, we use a function `getResultsUrl` to retrieve a part of the redirection url.

However, `getResultsUrl` is returning `undefined` because the switch never hits anything. We're checking on `component._self?.schema.ref`, which (in my test case) is `"https://opencatalogi.nl/oc.component.schema.json"`.

PR #218, opened 2023-01-31
feature/OC-523/ResultTemplates-use-correct-data

remko48:


pwa/src/templates/templateParts/resultsTemplates/cards/CardsResultTemplate.tsx
```diff
@@ -18,13 +18,11 @@ export const CardsResultTemplate: React.FC<CardsResultTemplateProps> = ({ compon
     return component._self;
   });
 
-  const { length } = _components;
-  const id = length + 1;
-  const _com = _components.every((el) => el._self.schema.ref !== "https://opencatalogi.nl/component.schema.json");
+  const _com = _components.every((el) => !el._self.schema.ref.includes("component.schema.json"));
```
lencodes, line 21: I find this piece of logic hard to understand.

Using more descriptive names, would be useful here. But explaining in a small comment _what_ you're actually doing might also help.

Also, does this need to be step by step, so:

```
if check a
  thereafter if check b
    thereafter if check c

=>

if (a && b && c) <= would this work?
```

You don't seem to set anything. So I'd expect the following to work as well:

```
const noResultFound = _components.every((component) => {
    return (
      !component._self.schema.ref.includes("component.schema.json") &&
      !component._self.schema.ref.includes("organisation.schema.json") &&
      !component._self.schema.ref.includes("application.schema.json")
    );
  });

  if (noResultFound) return <LeadParagraph>Geen Organisaties, Componenten of Applicaties gevonden.</LeadParagraph>;
```

pwa/src/templates/templateParts/resultsTemplates/table/TableResultTemplate.tsx
```diff
@@ -56,21 +56,31 @@ export const TableResultTemplate: React.FC<LayersResultTemplateProps> = ({ compo
...
+                      _.camelCase(
+                        t(
+                          `${
+                            (component._self.schema.ref.includes("component.schema.json") &&
+                              component.embedded?.nl?.embedded?.commonground?.layerType) ??
+                            "Unknown"
+                          } layer`,
+                        ),
+                      )
+                    ],
```
lencodes, line 69: I think this is getting too complex to be _within_ the TSX return statement. 

I'd opt to map the components _before_ getting to the return statement.

lencodes (changes requested):

PR #243, opened 2023-03-28
feature/OC-608/link-that-looks-like-a-button

britt11654:


pwa/static/themes/utrecht.css
```diff
@@ -203,4 +203,10 @@
   --web-app-alert-info-action-button-background-color: hsl(207 80% 35%);
   --web-app-alert-info-action-button-background-color-hover: hsl(207 80% 30%);
   --web-app-alert-info-icon-color: hsl(207 80% 35%);
+
+  /* 
+  * Utrecht component overrides
+  */
+  --web-app-button-secondary-action-background-color: inherit;
+  --web-app-button-secondary-action-color: inherit;
```
lencodes, line 211: _Feedback also goes for the `rotterdam.css` file_

I think we should specify that these tokens are meant _only_ for the button _link_ variant. This name suggests that its the `..-color` and `..-background-color` for the button itself which is (or might) not (be) true.

In the `component-overrides`, we're wrapping the overwrite in `.utrecht-button-link--secondary-action`, making it explicitly _only_ relevant in the `-button-link...`.

I think the simplest solution would be to rename these `--web-app` tokens to (e.g.): `--web-app-button-link-....`.

pwa/src/templates/templateParts/landingDisplayTemplates/categories/CategoriesLandingDisplayTemplate.tsx
```diff
@@ -110,9 +111,9 @@ export const CategoriesLandingDisplayTemplate: React.FC<CategoriesLandingDisplay
         ))}
       </div>
 
-      <Button icon={<ArrowRightIcon />} iconAlign="start" onClick={() => navigate("/categories")}>
-        Bekijk alle categorieën
-      </Button>
+      <ButtonLink onClick={() => navigate("/categories")} appearance="secondary-action-button">
+        <ArrowRightIcon /> Bekijk alle categorieën
+      </ButtonLink>
```
lencodes, line 116: This works as intended. However, I'm thinking: should we be passing the `href` prop for accessibility? 

I'm unsure, maybe check this with @Robbert.

pwa/src/templates/templateParts/searchComponent/SearchComponentTemplate.tsx
```diff
@@ -59,12 +60,15 @@ export const SearchComponentTemplate: React.FC<SearchComponentTemplateProps> = (
...
+          <SearchIcon />
           {t("Search")}
         </Button>
-        <Button icon={<ArrowRightIcon />} iconAlign="start" onClick={clearFilters} variant="secondary-action">
+
+        <ButtonLink onClick={clearFilters} appearance="secondary-action-button">
+          <ArrowRightIcon />
           {t("View all components")}
-        </Button>
+        </ButtonLink>
```
lencodes, line 71: I think, due to this click having direct side effects, we shouldn't use this `ButtonLink` component here. Please check with @Robbert.

Note: the side effect here is that we're clearing state and only _thereafter_ navigating the user. Instead of _only_ navigating.

lencodes (changes requested):
Nice changes @britt11654  👍🏼 

I've left some comments; some of which I do not know the answer to, hopefully @Robbert can help with those.

pwa/static/themes/utrecht.css
Robbert: In the case of the ButtonLink from Utrecht, I think using `web-app-button` actually makes sense, even for `button-link`: the `ButtonLink` component uses the tokens from `Button`, because they  are intended to look identical.

So I think the token names are probably fine the way they are now.

pwa/src/templates/templateParts/searchComponent/SearchComponentTemplate.tsx
Robbert: Probably this approach is already better than it was: previously the link could not be found using a screen reader.

@lencodes Maybe the ideal way method is using the [GatsbyLink state]() property, then the code would look something like this:

```jsx
<GatsbyLink to="/components" state={{
      resultDisplayLayout: filters.resultDisplayLayout,
      dependenciesDisplayLayout: filters.dependenciesDisplayLayout,
      landingDisplayLayout: filters.landingDisplayLayout,
      currentPage: filters.currentPage,
      applicationsCurrentPage: filters.applicationsCurrentPage,
    }} />
```

PR #256, opened 2023-04-11
Styling/oc 668/header

scar055:


pwa/src/templates/templateParts/header/HeaderTemplate.tsx
```diff
@@ -154,13 +154,21 @@ export const HeaderTemplate: React.FC<HeaderTemplateProps> = ({ layoutClassName
...
 
           <PrimaryTopNav
             mobileLogo={
               <div className={clsx(styles.logoContainer, styles.logoMobile)}>
-                <div onClick={() => navigate("/")} className={styles.organizationLogo} />
+                <Image
+                  alt="Logo van de Gemeente Rotterdam"
+                  src="images/LogoRotterdam.svg"
+                  className="utrecht-img--scale-down"
+                />
```
lencodes, line 171: By making this src attribute hardcoded (instead of setting it via the design token and css) the logo will _always_ be that of Gemeente Rotterdam.

Because of that, when changing to a new theme (e.g. Utrecht), the Rotterdam logo still shows, which is not intended.

pwa/src/templates/templateParts/header/HeaderTemplate.tsx
```diff
@@ -154,13 +154,21 @@ export const HeaderTemplate: React.FC<HeaderTemplateProps> = ({ layoutClassName
       <div className={styles.headerMiddleBar}>
         <Container layoutClassName={styles.primaryNavContainer}>
           <div className={clsx(styles.logoContainer, styles.logoDesktop)}>
-            <div onClick={() => navigate("/")} className={styles.organizationLogo} />
+            <Image
+              alt="Logo van de Gemeente Rotterdam"
+              src="images/LogoRotterdam.svg"
+              className="utrecht-img--scale-down"
+            />
```
lencodes, line 161: See other comment on the logo, same applies here.

pwa/src/templates/templateParts/header/HeaderTemplate.module.css
```diff
@@ -155,6 +149,7 @@ li .denhaag-link__icon {
   }
 
   .primaryNavContainer {
+    padding-block: 1em;
```
lencodes, line 152: I think we should use a design-token for this, if we do not have one, you can create one.

lencodes (changes requested):
Nice changes, I've left some comments.

lencodes:
By re-writing the logo styling (removing the design tokens logic), we unfortunately broke the other themes. Not much else was changed in this pull request. Therefore, I'm closing it (other small details will be implemented into the "verzamel" branch later on).

PR #263, opened 2023-04-20
♻️ Refactor/databadge/verbetering-databadge-component

Ridsko:


lencodes (changes requested):
Looks good!

One important note on this pull request though: making these changes breaks the color of the layer indicators (in multiple locations throughout the application: TableResultsTemplates, component detail templates, ...). 

E.g.: the DataBadges "Interactie" and "Services" should be the color of that layer (blue and green respectively). 

image


E.g.: all "Laag" badges that are _not_ "N.V.T." should have the color of the layer.

image

E.g.: the layer indicator here should also have the color of the layer.

image

PR #271, opened 2023-05-12
feat: add Amsterdam and Den Haag themes

Robbert:


lencodes (changes requested):
There's still too many inconsistencies in these themes to merge them, I think (see screenshot of the Den Haag theme).

Adding the packages themselves might be beneficial, however, actually adding the select options to the `themes[]` allows users to select this unfinished theme.

What do you think would be the best plan of action @Robbert?

image

PR #274, opened 2023-05-24
Refactor/radio button/component page

scar055:
radio buttons by coponent page de filter naar utrecht radio buttons veranderd


pwa/src/templates/templateParts/filters/verticalFilters/VerticalFiltersTemplate.tsx
```diff
@@ -415,10 +415,9 @@ export const VerticalFiltersTemplate: React.FC<VerticalFiltersTemplateProps> = (
                     onChange={() => setStatusRadioFilter(status.value)}
                     key={status.value}
                   >
-                    <input
-                      type="radio"
+                    <RadioButton
```
lencodes, line 418: The labels of the RadioButtons are not clickable anymore, which I think they should be. Also, theres no space between the radio button and the label.

pwa/src/templates/templateParts/filters/verticalFilters/VerticalFiltersTemplate.tsx
```diff
@@ -415,10 +415,9 @@ export const VerticalFiltersTemplate: React.FC<VerticalFiltersTemplateProps> = (
...
                       value={status.value}
-                      {...register("status")}
+                      // {...register("status")}
```
lencodes, line 420: This comment can be deleted altogether (we don't need to register this field). Same goes for the other RadioButtons.

lencodes (changes requested):

PR #292, opened 2023-06-22
feature/OC-729/Zwolle-theme

remko48:


pwa/static/themes/zwolle.css
```diff
@@ -0,0 +1,346 @@
...
+}
+
+/******* END OF THEME TEMPLATE ********/
+
+/*
+   * Theme-specific generics
+  */
+:root {
+  --zwolle-primary-blue: #476db7;
+}
```
lencodes, line 324: I would assume that there's a `--zwolle` token available for this, instead of making our own one and hardcoding the value?

lencodes (changes requested):
This works pretty great!

I did however, see some details which I think are worth addressing. 

1. Some tokens were not set correctly in the package we're including (I've shared a Google Docs with you, where we can collect these whenever we see them; maybe we can share it with Robbert from Frameless later on)

2. Because we're implementing the tokens from the packages now, we can clean up the tokens we've set manually (maybe you can create an issue in the Themes epic for this? And maybe we should do the cleanup _before_ adding all the other themes).

E.g. we're setting `--utrecht-link-hover-color: var(--web-app-color-link-active);` in component overrides, which we're manually setting in every theme (e.g. utrecht.css, rotterdam.css, etc.). However, if we just _remove_ this token altogether (from the overrides and the themes) the implemented tokens via npm will set them for us!

I noticed this, because the link hover was broken on Zwolle, by removing the logic altogether in all themes, the hover worked in every theme, making our own override redundant (and even breaking).

I can only assume that this goes for _many_ more tokens, so we can really simplify our own tokens this way.

lencodes (approved):

PR #295, opened 2023-06-22
Fixes deeplinking on applications

rjzondervan:


smisidjan (approved):

lencodes (changes requested):
I'm unsure if this fixed the issues. I've upped the frontend using docker, which serves the application using nginx.

The application is now served on: http://localhost:81/. 

Deep linking test:

- Deep linking to (e.g.) http://localhost:81/organizations/550cc92d-13fd-4692-a8c3-e670ff89c330 still gives the nginx 404;

Refreshing on a child-page:

- Refresh on (e.g.) http://localhost:81/components/ (note: incl. trailing slash!) works ✅ 
- Refresh on (e.g.) http://localhost:81/components (note: excl. trailing slash!) does not work, redirects me to http://localhost:8080/components/ 


I think these are the same issues we're running into on the live environment.

lencodes (approved):
After building the docker container again, this does work! 🥳

PR #297, opened 2023-06-26
feature/OC-735/organisation-overview

remko48:


pwa/src/templates/organizationsTemplate/OrganizationsTemplate.tsx
```diff
@@ -0,0 +1,96 @@
...
+import Skeleton from "react-loading-skeleton";
+import { GatsbyContext } from "../../context/gatsby";
+import { PaginatedItems } from "../../components/pagination/pagination";
+import ResultsDisplaySwitch from "../../components/resultsDisplaySwitch/ResultsDisplaySwitch";
+import { Heading } from "@utrecht/component-library-react/dist/css-module";
+import { useOrganization } from "../../hooks/organization";
+import { OrganizationHorizontalFiltersTemplate } from "../templateParts/filters/horizontalFilters/HorizontalFiltersTemplate";
+import { OrganizationDisplayTemplate } from "../templateParts/OrganizationDisplayTemplates/OrganizationDisplayTemplate";
+
+export const OrganizationsTemplate: React.FC = (props) => {
```
lencodes, line 16: We're not using `props`

pwa/src/templates/organizationsTemplate/OrganizationsTemplate.tsx
```diff
@@ -0,0 +1,96 @@
...
+    if (getOrganisations.isSuccess && screenSize === "mobile") {
+      setMarginPageDisplayed(2);
+    }
+    if (getOrganisations.isSuccess && screenSize === "mobile" && getOrganisations.data.pages > 100) {
+      setMarginPageDisplayed(1);
+    }
+    if (getOrganisations.isSuccess && screenSize !== "mobile") {
+      setMarginPageDisplayed(3);
+    }
+  }, [getOrganisations]);
```
lencodes, line 36: I'm thinking we might not need this in the implementation layer, couldn't `PaginatedItems` handle this within its own context? 

It knows how many pages it's showing.

pwa/src/templates/organizationsTemplate/OrganizationsTemplate.tsx
```diff
@@ -0,0 +1,96 @@
...
+                    getOrganisations.data.pages > 1000 ? styles.paginationLinkSmall : styles.paginationLink
+                  }
+                  nextClassName={
+                    getOrganisations.data.pages > 1000 ? styles.paginationLinkSmall : styles.paginationLink
+                  }
+                  activeClassName={
+                    getOrganisations.data.pages > 1000 ? styles.paginationActivePageSmall : styles.paginationActivePage
+                  }
+                  disabledClassName={styles.paginationDisabled}
+                  breakClassName={styles.breakLink}
```
lencodes, line 86: Do we actually need all those classNames being passed to `PaginatedItems`? 

I would expect `PaginatedItems` to have the ability to set them in its own context.

E.g.:

```
nextClassName={
  getOrganisations.data.pages > 1000 ? styles.paginationLinkSmall : styles.paginationLink
}
```

Everything's based on `.data.pages`, but we're also passing this to the component: `pages={getOrganisations.data.pages}`

lencodes: We're doing the same everywhere we're implementing this Pagination. I think the implementation is more complicated than it needs to be.

pwa/src/templates/organizationsTemplate/OrganizationsTemplate.module.css
```diff
@@ -0,0 +1,121 @@
...
+  gap: var(--web-app-size-lg);
+}
+
+.results {
+  flex: 3;
+
+  --utrecht-heading-4-font-weight: normal;
+  --utrecht-heading-4-font-size: var(--web-app-font-size-md);
+  --utrecht-heading-4-distanced-margin-block-end: var(--web-app-size-2xs);
+  --utrecht-heading-4-distanced-margin-block-start: var(--web-app-size-2xs);
```
lencodes, line 21: I don't think we're implementing a `heading-4` anywhere in this template. Is this doing anything?

lencodes, line 21: I think this is a copy paste from the components template, but some of this logic does not apply here.

E.g.: there's no two columns, so setting the flex: 3 1, and flex-direction is redundant here.

I'd suggest simply setting the width to 100%.

pwa/src/templates/templateParts/OrganizationDisplayTemplates/cards/CardsOrganizationDisplayTemplate.module.css
```diff
@@ -0,0 +1,19 @@
+.ComponentsGrid {
```
lencodes, line 1: `.componentsGrid`*

pwa/src/apiService/resources/organization.ts
```diff
@@ -17,11 +17,13 @@ export default class Organization {
...
   public getAll = async (filters: IFilters): Promise<any> => {
-    const { data } = await Send(
-      this._instance,
-      "GET",
-      `/organizations?page=${filters.currentPage}&limit=10&extend[]=all${filtersToQueryParams(filters)}`,
-    );
+    let url = `/organizations?page=${filters.organizationCurrentPage}&order[owns]=desc&limit=10&extend[]=all`;
+
+    if (filters.organizationSearch) {
+      url += `&_search=${filters.organizationSearch}`;
```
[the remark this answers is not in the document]
remko48: Backend is looking in to it. 
I don't know when they know the answer to this, so I will consider this out of scope for this pr.

lencodes (approved):
Very nice! Approved, but do first fix the following please.

I've noticed that the component loads twice, this was not the case before this pull request (I think; it might have something to do with the pagination changes). See the attached video.


https://github.com/OpenCatalogi/web-app/assets/66552932/183958ac-c537-4487-b8b0-fce0aba9392e

PR #299, opened 2023-06-26
Feature/oc 762/lead paragraph read more

lencodes:
Check commit messages and the issue (OC-762) for context


pwa/src/components/expandableLeadParagraph/ExpandableLeadParagraph.tsx
```diff
@@ -0,0 +1,46 @@
+import * as React from "react";
+import * as styles from "./ExpandableLeadParagraph.module.css";
+import { Button, Paragraph } from "@utrecht/component-library-react";
+import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
+import { faChevronDown } from "@fortawesome/free-solid-svg-icons";
```
remko48, line 5: We usually use `faChevronRight` for collapsibles so when the collapsible is closed the chevron faces to the right and when we open the collapsible we rotate it `90deg` so it faces downwards.

remko48 (approved):
Looks good,  just one small comment after that you can merge it 👍

PR #303, opened 2023-07-26
feature/OC-776/submit-component-fix

remko48:


pwa/src/templates/componentDetail/ComponentsDetailTemplate.tsx
```diff
@@ -72,6 +72,17 @@ export const ComponentsDetailTemplate: React.FC<ComponentsDetailTemplateProps> =
 
   const organisation = _getComponent?.data?.embedded?.url?.embedded?.organisation;
 
+  const isValidHttpUrl = (src: string) => {
```
lencodes, line 75: The name of this function suggest it to return a `boolean`, but it returns the source _if_ it's a valid one. Maybe something like this would be more descriptive:

```
const imageHasValidSource = (src: string): boolean => {
  try {
    const url = new URL(src);

    return url.protocol === "http:" || url.protocol === "https:"; // return url.protocol.contains("http"); might also work
  } catch () {
    return false;
  }
};
```

PR #307, opened 2023-08-08
Helm updates with resource limits and better documentation

rjzondervan:


WilcoLouwerse (approved):
Looks good, i would maybe mention somewhere all the way at the top of the readme that the Gateway + CoreBundle is required and that general documentation for this can be found at: Link to CoreBundle Readme

PR #327, opened 2023-09-20
Development to main

remko48:


pwa/src/templates/templateParts/resultsTemplates/table/TableResultTemplate.tsx
```diff
@@ -69,7 +71,7 @@ export const TableResultTemplate: React.FC<TableResultTemplateProps> = ({ compon
   };
 
   return (
-    <TableWrapper>
+    <TableWrapper touchScreen={screenSize === "tablet" || screenSize === "mobile"}>
```
lencodes, line 74: See my comment on the `TableWrapper` itself in the conduction/components pull request.

Would be nice if we could simply implement this like:

```
<HorizontalOverflowWrapper>
 ... content
</Horizon...>
```

lencodes (changes requested):
Nice changes @remko48 

Left some small comments.

PR #335, opened 2023-09-27
feature/OP-58/Link-NLDS

remko48:


lencodes (changes requested):
Very nice changes @remko48 looks a lot better than that it did.

I've left some inline comments. 

Also, I see we're adding quite a few classes in multiple files which are doing the same thing (`.icon, `.link`, `.linkText`). Especially the `.icon` we're setting 49 times!

I think what Frameless does in these instances is create a static className, e.g. `"link--icon"`. And then set the required styling somewhere in a global file. This way we only need to set the css rule once (making it much easier to change in the future).

Another possible solution could be to create a wrapper component around this link, which accepts some props (e.g. if we send a icon, the spacing of the icon and label will be handled in the component itself).

PR #343, opened 2023-10-02
Development to main, week 39

remko48:


pwa/src/styling/global.css
```diff
@@ -58,4 +63,25 @@ body {
   --web-app-layer-color-services: #69b090;
   --web-app-layer-color-data: #7a51c8;
   --web-app-layer-color-unknown: #586265;
+
+  --web-app-footer-logo-container-height: 50px;
```
lencodes, line 67: I'm unsure if it's worth to keep these tokens here. Some of them are used in the theme files as well (which I suppose is a temporary solution, since we want to discontinue these local theme files I think). 

Others we're not really using anywhere else (e.g. `--web-app-container-width-lg` is **only** used in `ComponentsTemplate` (other tokens as well). Maybe it would be better to simply just set them there with a hardcoded value, reducing complexity? What do you think?

lencodes (approved):
Looks good. Did leave a comment, maybe we can have a quick call to discuss?

PR #347, opened 2023-10-09
feature/OP-67/Header-NLDS

remko48:


pwa/src/services/isHomepage.ts
```diff
@@ -0,0 +1,12 @@
+//This function checks whether you are at the homepage considering a prefix value to unsure that code depending on the homepage works accordingly
```
lencodes, line 1: I think adding a note that its the prefix set in `config.js` would be a nice addition.

pwa/src/templates/templateParts/header/HeaderTemplate.tsx
```diff
@@ -222,66 +221,38 @@ export const HeaderTemplate: React.FC<HeaderTemplateProps> = ({ layoutClassName
...
+          <BreadcrumbNav className={styles.breadcrumbs} label={t("Breadcrumbs")}>
+            {translatedCrumbs.map((crumb: any, idx: number) => {
+              if (crumbs.length !== idx + 1) {
+                return (
+                  <React.Fragment key={idx}>
+                    <BreadcrumbNavLink
+                      className={styles.breadcrumbNavLink}
+                      onClick={(e: any) => handleBreadcrumbClick(e, crumb.pathname)}
+                      href=""
+                    >
```
lencodes, line 235: Might be nice to make a local component, holding the breadcrumbs, which itself keeps track on if it should render, like so (dummy code):

```
const Breadcrumbs: React.FC<{pathname: string}> = ({pathname}) => {
  if (isHomepage(pathname) return;

  return ... // component like we had it
}
```

Makes the return statement a bit easier to read and pulls the logic apart.

pwa/src/templates/templateParts/header/HeaderTemplate.tsx
```diff
@@ -222,66 +221,38 @@ export const HeaderTemplate: React.FC<HeaderTemplateProps> = ({ layoutClassName
...
-                        onClick={(e: any) => handleBreadcrumbClick(e, crumb.pathname)}
-                        href=""
-                      >
-                        {crumb.crumbLabel}
-                      </BreadcrumbNavLink>
-                    );
-                  }
-                  return (
-                    <BreadcrumbNavLink key={idx} className={styles.breadcrumbDisabled} current disabled href="">
+      {isHomepage(pathname) && (
```
lencodes, line 224: I think we want to show the breadcrumbs when it's **not** the homepage. Right now, reading this, it would suggest that we're showing it when it **is** the homepage.

I would expect `!isHomepage ...`

lencodes (changes requested):
Nice changes 👍🏼 

Left some comments.

PR #355, opened 2023-10-20
Development to main, week 42

remko48:


pwa/src/templates/templateParts/footer/FooterTemplate.tsx
```diff
@@ -1,273 +1,299 @@
...
+
+  return (
+    <section>
+      <Heading3 className={styles.dynamicSectionTitle}>{t(content.title)}</Heading3>
+
+      {content.items.map((item, idx) => (
+        <div key={idx} className={styles.dynamicSectionContent}>
+          {item.label && <strong>{t(item.label)}</strong>}
+          {/* External Link */}
+          {item.link && item.link.includes("http") && (
```
lencodes, line 99: I suggest making four local components, for the links. Making the dynamic return statement much easier to read.

Resulting in (e.g.)

```
const DynamicSection: React.FC<{ content: TDynamicContentItem }> = ({ content }) => {
  const [, setFilters] = React.useContext(FiltersContext);

  return (
    <section>
      <Heading3 className={styles.dynamicSectionTitle}>{t(content.title)}</Heading3>

      {content.items.map((item, idx) => (
        <div key={idx} className={styles.dynamicSectionContent}>
          {item.label && <strong>{t(item.label)}</strong>}

          {item.link && item.link.includes("http") && <ExternalLink />}

          {item.link && !item.link.includes("http") && <InternalLink />}

          {item.markdownLink && !item.setFilter && <MarkdownLink />}

          {!item.link && item.setFilter && <FilterLink />}

          {!item.knk && !item.setFilters && !item.markdownLink && <NoLink />}
        </div>
      ))}
    </section>
  );
};
```

pwa/static/.env.development
```diff
@@ -14,5 +14,11 @@ GATSBY_ADMIN_DASHBOARD_URL=https://admin.opencatalogi.nl
...
 GATSBY_NL_DESIGN_THEME_CLASSNAME=rotterdam-theme
 # GATSBY_GITHUB_ORGANIZATION_URL=""
+
+# Header
 GATSBY_HEADER_LOGO_URL=https://www.rotterdam.nl/images/logo-base.svg
 
+# Footer
+GATSBY_FOOTER_SHOW_CREATOR="false"
+GATSBY_FOOTER_LOGO_URL="false"
+GATSBY_FOOTER_CONTENT="https://raw.githubusercontent.com/OpenCatalogi/web-app/25995205e1dbb043822d1c58c5c23f8e6f77ff7e/pwa/src/templates/templateParts/footer/FooterContent.json"
```
lencodes, line 24: **Very nice!**

lencodes (changes requested):
Nice changes @remko48, I've left some comments. If you have any questions, we can discuss in a call. 

I think we should move the ParsedHTML logic to the `@conduction/components` package, instead of duplicating it here. This would allow us to easily implement it in other applications and would be easier to manage.

(Also the current implementation in web-app creates many IDE warnings due to its settings).

PR #357, opened 2023-10-24
Feature/op 81/query param filters

lencodes:
~~Before actually implementing the query params, I thought it'd be better to first implement our new filters way of working (open catalogi was still using older methods).~~

~~Also before implementing the query params, I'd like to merge this to minimise the code changes per pull request.~~

Forgot to merge this timely, but the additional changes were smaller than expected, so should be okay to keep in one pull request 🤞🏼 

**This pull request includes:**

- Refactored all context in the application, to be more alike the context in other projects (custom hook per context-type, with one Provider, instead of multiple);
- Added a `filtersToUrlQueryParams` service, which rewrites filters to URL Query params (with URIEncoding, so that we can keep spaces);
- Added a `getParsedParamsFromSearch` service, which creates a JSON object from the query params (incl. decoding of numbers, arrays, booleans and the encoded strings);
- Refactored the FilterContext to push all changes to the URL and thereafter update the filters (this way the URL is always leading, making deep-linking possible)

**Known issue / to-do:**

- The clearing of filters in the `ActiveFiltersTemplate` doesn't work, unsure why at this moment
- On the Component page, when you select (e.g.) Cards as the display and refresh the page, it switches back to Table (default) before going back to Cards. The result is what we want, but the switching should be addressed still.


remko48 (approved):
Nice!

PR #375, opened 2023-10-31
feature/OP-108/pagination-limit

remko48:


pwa/src/components/paginationLimitSelect/PaginationLimitSelect.module.css
```diff
@@ -0,0 +1,12 @@
+.container {
+  display: flex;
+  flex-wrap: wrap;
+  gap: 8px;
```
lencodes, line 4: I think we should use a token here.

pwa/src/templates/templateParts/footer/FooterTemplate.tsx
```diff
@@ -129,7 +127,7 @@ const Logo: React.FC = () => {
 };
 
 const WithLoveByConduction: React.FC = () => {
-  if (process.env.GATSBY_FOOTER_SHOW_CREATOR === "false") return <></>;
+  // if (process.env.GATSBY_FOOTER_SHOW_CREATOR === "false") return <></>;
```
lencodes, line 130: Was this intentional?

lencodes (changes requested):
Very nice! Left some inline comments, also have a few other global comments.

1. Display is not set automatically (unsure if that changed in this pull request, but it's there), see screenshot; because of this the results don't actually show:

image

2. When setting a higher amount of results, there's a big enough scroll to completely remove the filters view in the components template, see screenshot:

image

I think we should make the filters sticky, keeping them in view when scrolling.

PR #376, opened 2023-10-31
feature/OP-108/current-page

remko48:


pwa/src/context/pagination.ts
```diff
@@ -0,0 +1,25 @@
...
+import { GlobalContext } from "./global";
+
+export interface IPaginationContext {
+  currentPage: number;
+  organizationCurrentPage: number;
+  applicationCurrentPage: number;
+}
+
+export const defaultPaginationContext: IPaginationContext = {
+  currentPage: 1,
```
lencodes, line 11: I think `componentsCurrentPage` would be a better name for this.

lencodes (changes requested):
Nice changes.

I see a few issues with the queryParams that are happening: 

My current page is 91, but in the URL it's 1. When navigating to currentPage=91 it DOES actually go to the right page, but the URL is reset to 1.
image

Organisations and Applications don't show their filters in the queryParams at all, making deep linking impossible. 
image

Also, I see that the pagination also shows when there are no results available:
image

lencodes (approved):
Nice 👍🏼

PR #383, opened 2023-11-06
feature/OP-64/header-refactor

remko48:


pwa/src/hooks/useHeaderTopNavItems.ts
```diff
@@ -0,0 +1,129 @@
...
+import { navigate } from "gatsby";
+import { useTranslation } from "react-i18next";
+
+type THeaderTopNavItem = {
+  label: string;
+  type: "readme" | "internal" | "external";
+  current: {
+    pathname: string;
+    filterCondition?: {
+      filter: string;
```
lencodes, line 13: I think `filterKey` might be a more descriptive name.

PR #385, opened 2023-11-07
REFACTOR ComponentDetail template to show actual data

lencodes:
This pull requests rewrites the "Extra informatie" section on the component detail page to show actual data. Note: some of the fields were 100% hardcoded, and never showed any data. I could not find the corresponding data for some of these fields. The ones I couldn't find, I simply removed. I also added new ones I _did_ find in the documentation.

### DIFF
--- a/pwa/src/templates/componentDetail/ComponentsDetailTemplate.tsx
+++ b/pwa/src/templates/componentDetail/ComponentsDetailTemplate.tsx
@@ -72,6 +72,9 @@ export const ComponentsDetailTemplate: React.FC<ComponentsDetailTemplateProps> =
       });
     });
 
+  const gemma = _getComponent.data?.embedded?.nl?.embedded?.gemma;
+  const legal = _getComponent.data?.embedded?.legal;
+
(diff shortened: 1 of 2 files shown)


pwa/src/templates/componentDetail/ComponentsDetailTemplate.tsx
```diff
@@ -492,15 +495,53 @@ export const ComponentsDetailTemplate: React.FC<ComponentsDetailTemplateProps> =
 
             <Table>
```
remko48, line 496: I think we should let the user know that there is no information instead of a blank page.
image

remko48 (approved):
Just one small comment, otherwise approved!

pwa/src/templates/componentDetail/ComponentsDetailTemplate.tsx
lencodes: Thanks, missed that scenario in my own tests. I've completely removed the section, _if_ there's no extra information available. 

Note: I still need to test the logic, but I cannot due to the gateway not returning components at the moment.

PR #390, opened 2023-11-07
bug-fixes

remko48:


lencodes (approved):
Approved.

Do think we should fix the underlying issue with the `"readme" | "markdown"`. I understand we now need both, but we should remove the `"readme"` from the codebase _and_ the configuration everywhere.

PR #403, opened 2023-11-24
feature/OP-112/download

remko48:


pwa/src/templates/templateParts/download/DownloadTemplate.tsx
```diff
@@ -1,63 +1,80 @@
...
+              toggleVisibility(label ?? downloadLink);
             }}
-            layoutClassName={styles.popup}
           />
-        </div>
-      )}
+
+          {isVisible[label ?? downloadLink]?.value && (
+            <div className={styles.overlay}>
+              <NotificationPopUp
```
lencodes, line 54: I think we should refactor the placement, and animation, of this pop-up. Check out the Rating pop-up, this has already been updated.

Note: I now see that the Rating pop-up does not always center correctly, maybe you can fix that in this pull request (if it's easy to fix).

pwa/src/templates/templateParts/download/DownloadTemplate.tsx
```diff
@@ -1,63 +1,80 @@
...
-              icon: <FontAwesomeIcon icon={faArrowLeft} />,
-              handleClick: () => ({}),
+      {items.map(({ label, size, type, downloadLink }, idx) => (
+        <React.Fragment key={idx}>
+          <DownloadCard
+            label={label ?? downloadLink}
+            type={type ?? downloadLink.substring(downloadLink.lastIndexOf(".") + 1)}
+            {...{ size, icon }}
+            handleClick={() => {
+              toggleVisibility(label ?? downloadLink);
```
lencodes, line 48: I'm unsure if defaulting to the `downloadLink` is the right thing to do for a label, see screenshot. Is there any other way we can actually find the file name perhaps? This file in particular is called `ZGWBundle-main`, which would be a good name to use (maybe something the backend can implement).

Otherwise, I think simply "No name available" might be better than showing the URL. 

image

pwa/src/templates/componentDetail/ComponentsDetailTemplate.tsx
```diff
@@ -79,9 +77,17 @@ export const ComponentsDetailTemplate: React.FC<ComponentsDetailTemplateProps> =
   const _categories =
     layer &&
     _getComponent.data?.categories.map((category: any) => {
-      return categories[layer]?.find((_category) => {
+      const result = categories[layer]?.find((_category) => {
         return _category.value === category;
       });
+
+      if (result === undefined) {
```
lencodes, line 84: I think `if (!result)` would do the same. Or if `result === false` or something should also hit the `else`, than maybe not.

lencodes (changes requested):
Nice changes. I've left some comments.

PR #405, opened 2023-11-29
Development to main, week 48

remko48:


pwa/src/templates/componentDetail/ComponentsDetailTemplate.tsx
```diff
@@ -49,18 +48,17 @@ import { GitHubLogo } from "../../assets/svgs/GitHub";
 import { DependenciesTemplate } from "../templateParts/dependenciesTemplates/ComponentDependenciesTemplate";
 import { useResultDisplayLayoutContext } from "../../context/resultDisplayLayout";
 import { ComponentCardsAccordionTemplate } from "../templateParts/componentCardsAccordion/ComponentCardsAccordionTemplate";
-import { DownloadTemplate } from "../templateParts/download/DownloadTemplate";
+import { DownloadProps, DownloadTemplate } from "../templateParts/download/DownloadTemplate";
```
lencodes, line 51: Usually we use the `...Props` keyword only for the props we're sending to a component. E.g. int he `DownloadTemplate` we have the `DownloadTemplateProps`, which are its props.

Now, we've added a `DownloadProps` interface which is actually not the main props for this component which might be confusing. I'd opt for making it a type, and naming it something along the lines of:

```
export type TDownloadableItem = { ... structure of type here }
```

pwa/src/templates/templateParts/download/DownloadTemplate.module.css
```diff
@@ -3,9 +3,10 @@
   max-width: var(--web-app-download-popup-width);
   top: 50%;
   left: 50%;
-  transform: translate(-50%, -50%);
   z-index: 700;
-  outline: 0;
+  margin: 0 auto;
+  margin-left: -250px;
+  margin-top: -250px;
```
lencodes, line 9: Nice. I do think however, that we should make these margins relative to the `--web-app-download-popup-width`. You're already doing it, but implicitly. 

Making it like this, ensures that when the width is of the pop-up is changed, its still being centered;

```
margin-left: calc(var(--web-app-download-popup-width) / 2 * -1);
margin-top: calc(var(--web-app-download-popup-width) / 2 * -1);
```

lencodes: Edit: the rules could be `margin-inline-start` and `margin-block-start` instead of left and top 👍🏼

pwa/src/templates/templateParts/download/DownloadTemplate.tsx
```diff
@@ -1,63 +1,82 @@
...
 import { useTranslation } from "react-i18next";
-import clsx from "clsx";
 import { useGatsbyContext } from "../../../context/gatsby";
+import { TOOLTIP_ID } from "../../../layout/Layout";
 
 interface DownloadTemplateProps {
-  icon: JSX.Element;
-  label: string;
-  sizeKb: string;
+  items: DownloadProps[];
```
lencodes, line 12: See the comment about the `TDownloadableItem`.

pwa/src/templates/templateParts/download/DownloadTemplate.tsx
```diff
@@ -1,63 +1,82 @@
...
+  backUrl: string;
+  icon?: JSX.Element;
```
lencodes, line 14: In some implementations, we're passing a `FontAwesomeIcon` type (unsure the exact type), instead of a `JSX.Element`. I'd do that here as well (right now, we can add **anything** to this `icon`, which is prone to error).

pwa/src/templates/componentDetail/ComponentsDetailTemplate.tsx
```diff
@@ -121,6 +127,13 @@ export const ComponentsDetailTemplate: React.FC<ComponentsDetailTemplateProps> =
     },
   ];
 
+  const downloads: DownloadProps[] = _getComponent?.data?.embedded?.downloads?.map((download: any) => ({
+    label: download.naam,
+    size: download.grootte,
+    type: download.type,
+    downloadLink: download.url,
+  }));
```
lencodes, line 135: I'm thinking it might be better to simply send the `download` object (from `...embedded.downloads`) to the `DownloadTemplate`.

pwa/src/templates/templateParts/download/DownloadTemplate.tsx
lencodes: I'd call this 'downloads' instead of `items`.

pwa/src/templates/templateParts/download/DownloadTemplate.tsx
```diff
@@ -1,63 +1,82 @@
...
+                )}
+                primaryButton={{
+                  label: "Download",
+                  icon: <FontAwesomeIcon icon={faDownload} />,
+                  handleClick: () => open(downloadLink),
+                }}
+                secondaryButton={{
+                  label: t("Go back"),
+                  icon: <FontAwesomeIcon icon={faArrowLeft} />,
+                  href: `${backUrl}`,
```
lencodes, line 71: `href: backUrl` would do the same.

pwa/src/templates/templateParts/download/DownloadTemplate.tsx
```diff
@@ -1,63 +1,82 @@
...
 
-  const NotificationPopUpController = _NotificationPopUp.controller;
-  const NotificationPopUp = _NotificationPopUp.NotificationPopUp;
+  const toggleVisibility = (label: string) => {
+    setIsVisible({
+      [label]: {
+        value: true,
+      },
+    });
+  };
```
lencodes, line 36: I think we can simplify this `isVisible` logic. Right now, we're not really typing it, allowing `any`.

Eventually we're only checking if the `isVisible[label] = true`, what if we simply checked if this label is the current visible one?

Like so:

```
const [isVisible, setIsVisible] = React.useState<string>("");
const toggleVisibility = (label: string) => setIsVisible(label);


// The check can now become
isVisible === (label ?? downloadLink);

// The hide can become
setIsVisible("");
```

lencodes: Iterating on the idea above, we could also use the index of the given array, like so:

```
const [visibleItemIdx, setVisibleItemIdx] = React.useState<number>(-1); // default to -1, which is never available

<DownloadCard
  ... other props
  handleClick={() => setVisibleItemIdx(idx)} // instead of the toggle, simply call the state directly
/>

...

{visibleItemIdx === idx && ( // check once if the pop-up should exist
  <div className={styles.overlay}>
    <NotificationPopUp
      isVisible // do not check here again
      hide={() => setVisibleItemIdx(-1)} // set to -1 to hide everything
      ... nothing changed
    />
  </div>
)}
```

lencodes (changes requested):
Nice changes @remko48 

I've left some comments.

PR #411, opened 2023-12-06
Update publiccode.yaml

RonaldvCortenberghe:
layertype interface (laag 5) toegevoegd


publiccode.yaml
```diff
@@ -44,6 +44,8 @@ dependsOn:
     - versionMin: 1.0.0
     
 nl:
+  commonground:
+  - layerType: "interface"
```
remko48, line 48: This line can be replaced with `layerType: interface` the reason for that is:
- The `-` suggests that layerType is part of an array which it is not suppose to be
- There is no need for the value to be put into quotations

remko48 (changes requested):
Thanks @RonaldvCortenberghe for adding the layerType.
I did request some changes to be done but we can also do that for you if you want.

remko48 (approved):

PR #469, opened 2024-04-29
feature/OP-278/missing-charts

remko48:


pwa/src/templates/charts/CommongroundChartsTemplate.tsx
```diff
@@ -20,47 +20,33 @@ interface dataProps {
...
+interface hoverProps {
+  ontwikkelingsfases?: number | undefined;
+  initiatieven?: number | undefined;
+  domein?: number | undefined;
+  layer?: number | undefined;
+  rating?: number | undefined;
+  organization?: number | undefined;
+}
+
 export const CommongroundCartsTemplate: React.FC = () => {
```
bbrands02, line 32: Shouldn't this be "Charts"

remko48: Nice it should thanks 👍

